PR #20, opened 2026-04-20
fix[kafka streams skill]: update approach to deserialization

Cerchie:
## Description

For context, I ran into this while reviewing the skill recently, Claude's explanation of the error I ran into: 

```
 kafka-avro-console-consumer defaults 
  to KafkaAvroDeserializer for both key
   and value. Our topology writes      
  string keys with Serdes.String() (no
  SR framing), so the Avro deserializer
   chokes on the first byte — it     
  expects a 0x00 magic byte followed by
   a 4-byte schema ID, and a raw UTF-8
  region name doesn't have that.

```

This should fix it so the generated deserializer won't choke.


davetroiano (approved):
lgtm!

PR #22, opened 2026-04-27
kafka streams: update instructions

Cerchie:
## Description

I started a branch to address the Kafka Streams compilation error issue we ran into while doing the bug bash. However, the more times I gave it a manual run, the more new compilation and build errors I ran into (Avro type issues, maven mismatches). So, I decided to fix this at the instruction level instead and ask the agent to self-verify as well if possible so there's a chance these kinds of things can be fixed before reaching the user.

The changes in the other files are related to helping the evals pass at a higher percentage, I noticed they were squeaking by at 89-90%. 

## Testing

## Review

- [x] DTX/DevRel reviewer assigned


davetroiano (approved):
lgtm! I tested this a few times and didn't hit any build errors

davetroiano (approved):
lgtm! I tested this a few times and didn't hit any build errors

PR #12, opened 2026-04-03
task: python client skill updates

Cerchie:
## Description

Adds JSON schema registry instead of Avro, plus a couple fixes in the consumer and docker ref files to make things run more smoothly on local. 

## Skill Quality Checklist

- [x] Skill [description is optimized](https://agentskills.io/skill-creation/optimizing-descriptions) for accurate triggering
- [x] Uses concrete examples; avoids verbose paragraphs
- [x] Feedback loops like unit tests are built-in to enable agent iteration

## Testing

- [x] Evals pass at 90%+ threshold

## Review

- [ ] SME reviewer identified: N/A for this one
- [x] DTX/DevRel reviewer assigned

PR #6, opened 2026-03-30
feat: add confluent-cloud-cdc-tableflow skill

leeec1993:
## Description

Adds a new skill for setting up end-to-end Change Data Capture (CDC) pipelines on Confluent Cloud using Debezium source connectors, Flink for transformation, and Tableflow for data lake integration (Iceberg/Delta Lake).

### Supported Databases
- PostgreSQL CDC Source V2
- MySQL CDC Source V2
- SQL Server CDC Source V2
- Oracle XStream CDC Source
- DynamoDB CDC Source

### Skill Structure
- `SKILL.md` — Main skill definition with 5 workflow phases (MCP validation, discovery, planning, execution, verification)
- `evals/evals.json` — 7 evaluation test cases covering all supported databases and troubleshooting
- `references/` — 4 reference docs (connector configs, database prerequisites, Flink SQL patterns, troubleshooting)

### Key Features
- Prefers Confluent MCP Server tools with CLI fallback
- Auto-reads MCP `.env` file for environment/cluster defaults
- Handles Debezium type conversions in Flink SQL
- Supports both managed and BYOB Tableflow storage

## Skill Quality Checklist

- [ ] Skill name uses [gerund form](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#naming-conventions) (e.g., `creating-connectors`, not `create-connector`)
- [x] Skill [description is optimized](https://agentskills.io/skill-creation/optimizing-descriptions) for accurate triggering
- [x] Uses concrete examples; avoids verbose paragraphs
- [x] Feedback loops like unit tests are built-in to enable agent iteration

## Testing

- [ ] Evals pass at 90%+ threshold

## Review

- [ ] SME reviewer identified:
- [ ] DTX/DevRel reviewer assigned

## Additional Context

Pipeline flow: **Database → Debezium CDC Connector → Kafka + Schema Registry → Flink (decode & transform) → Tableflow → Iceberg/Delta Tables**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4, opened 2026-03-24
Add local review dashboard for viewing token burn per session

Cerchie:
Adds a review dashboard for reviewing skills' token burn locally. This will allow us to see how much a skill burns per session when we manually review a skill. 
Screenshot 2026-03-27 at 10 21 03 AM

How does it work? Claude logs token usage to files on your machine.
This script grabs usage by session, sums it up, and displays it on a dashboard. Note that prompt-cached tokens are not charged full-price, but 1/4-1/10 of the price. 

> [!NOTE]
> Now, some of these tokens were duplicated originally, so we only count usage blocks where there was a stop reason, [ensuring deduplication](https://github.com/confluentinc/agent-skills/pull/4/changes#diff-b25d473a56bfcd7cb98e1b15d528641b6f808235c2ad4535822bf9d13477084fR88).

## How to clicktest: 

1. Check this branch out. You can `cd ~/.claude && code .` (if you have the 'code' shortcut established, otherwise, find another way to open it in your IDE) to view your sessions in your IDE: 
Screenshot 2026-03-27 at 9 57 57 AM

2.  Download the attached verification script: [export_tokens_csv.py](https://github.com/user-attachments/files/26310848/export_tokens_csv.py) Grab a session id and run it like: 
```
./tools/skill_review_dashboard.py --session-id [YOUR_SESSION_ID]
```

3. Now keep that dashboard up. Run `./export_tokens_csv.py --session-id [YOUR_SESSION_ID]`
4. Verify that when you manually (or use a CSV shortcut) add the columns _with a value for the stop reason_ that they add up to what the dashboard says:
Screenshot 2026-03-27 at 9 56 16 AM

PR #25, opened 2026-05-04
Add msk-migration skill

amorgan-cflt:
> **Note:** This PR replaces #24, which has been closed. The original branch had a 15 MB customer-derived state file in its git history that was slowing down clones; this branch is opened fresh from `main` with a single commit and no large file in history. Substantive content matches the final state of #24 after DTX feedback was incorporated.

## Description

Adds a new skill for AWS MSK to Confluent Cloud migration assessment and planning.

**Use case:** A frequent customer ask in Confluent's Kafka takeout strategy. The skill helps customers (or AI agents acting on their behalf) audit their MSK environment via KCP scans or manual intake, identify migration risks early, and produce a structured Plan that downstream execution stages will consume.

**V0.1 MVP scope:** Assess and Plan stages.
- **Assess** — runs a 16-row Red Flags audit on a KCP state file or manual profile, produces an Environment Summary grounded in evidence from the source, scope-checks before the Plan stage.
- **Plan** — produces a 12-section Plan document with cluster type, sizing (P95 + peak), networking decision, auth approach, switchover approach, pre-migration workstream, risks, open questions, and next step. Tables follow standardized formats; every number carries an inline citation.

V0.2 will extend to Provision / Migrate / Switchover / Monitor following early DTX feedback. The skill is the assessment-and-planning bridge — it does not execute provisioning, data migration, or cutover today.

**What ships:**

```
skills/msk-migration/
├── README.md
├── SKILL.md                          # Dispatcher (~240 lines)
├── references/
│   ├── assess.md                     # Assess stage — 16-row Red Flags audit, scan coverage
│   ├── plan.md                       # Plan stage — 12-section template, sizing math, decision tables
│   ├── kcp-commands.md               # KCP CLI command reference
│   └── mcp-integration.md            # Local Confluent MCP guardrails
├── assets/
│   └── migration-profile.yaml        # Manual intake template (when KCP isn't available)
└── evals/
    ├── evals.json                    # 14 behavioral test scenarios
    ├── human-review-checklist.md     # Items the agentskills.io spec places in human review (full contents inlined below)
    └── mock-environments/            # Fixture profiles (manual YAML + small serverless state file)
```

## Skill Quality Checklist

- [x] Skill [description is optimized](https://agentskills.io/skill-creation/optimizing-descriptions) for accurate triggering — explicit MSK-only scope, named negative triggers (non-MSK Kafka sources, greenfield projects), concrete examples of triggering intent
- [x] Feedback loops like unit tests are built-in to enable agent iteration — 14-eval test suite at `evals/evals.json` with structured assertions and mock fixtures, runnable via the agentskills.io eval-driven loop

## Testing

**Final state: 14 of 14 evals pass with the skill loaded.** Validated through 9 iterations of an automated eval-driven loop on an 18-eval suite following the [agentskills.io spec](https://agentskills.io/skill-creation/evaluating-skills); after early-review feedback, 4 state-file-path evals + their large fixture were scoped out of V0.1 for fixture-quality reasons (covered below) and the remaining 14 evals all pass. With-skill pass rate: 100% mean. Without-skill baseline: ~37%. Delta: ~+63pp.

### Eval scenarios

The 14 V0.1 evals are organized into three groups: KCP state-file path (1, small fixture), manual `migration-profile.yaml` path (7), and conceptual / behavior scenarios (6). Four additional state-file-path evals are scoped to V0.2 — see "What's deferred to V0.2" below.

**KCP state-file path (1 eval, 24 KB fixture):**

- `eval-12 assess-serverless` — small KCP state file for an MSK Serverless cluster; tests the serverless-specific ACL-null disambiguator (KCP intentionally skips the Kafka Admin scan on serverless — null is expected, not a scan gap).

**Manual `migration-profile.yaml` path (7 evals):**

- `eval-03 dedicated-large-fleet` — manual profile describes a large MSK Provisioned fleet; tests Dedicated escalation only on hard-limit triggers (specifically eCKU cap), with live-doc routing for current caps.
- `eval-05 manual-no-kcp` — user explicitly can't run KCP; tests the 11-question intake walkthrough, recording "I don't know" as a known gap rather than fabricating.
- `eval-14 assess-iam-private-manual` — IAM-private clusters via manual profile; tests Red Flags audit + manual-profile null-vs-empty semantics on `acl_count: null`.
- `eval-15 hidden-connectors-manual` — hidden Connect activity via manual profile; tests Row 7/Row 15 cross-reference where the connector scan never ran.
- `eval-16 full-plan-open-questions-manual` — full Plan from a manual profile with auth and SR as open questions; tests the 12-section Plan template with manual-profile gap annotation.
- `eval-17 assess-serverless-manual` — MSK Serverless via manual profile; tests recognition of serverless from manual data and the right ACL-null framing.
- `eval-18 assess-hidden-cluster-manual` — Row 16 cost-to-inventory reconciliation via manual profile (`cost_line_items`); tests reconciliation on manual data + the manual-intake disambiguator for absent fields.

**Conceptual / behavior (6 evals):**

- `eval-01 kcp-fresh-start` — user starting from scratch wants to use KCP; tests that the skill introduces KCP, presents the 6-step scan sequence, and asks before running any commands.
- `eval-04 intake-triage` — generic migration intent ("we want to migrate off MSK") with no stage signal; tests that the skill triages without fabricating environment details or jumping to a specific intake path.
- `eval-08 out-of-scope-baremetal` — user has self-managed Kafka on bare metal; tests scope enforcement (skill declines, redirects to docs and the Confluent account team without fabricating coverage).
- `eval-09 mtls-azure-availability` — user has mTLS auth and wants to land on Azure; tests live-doc routing for feature availability without making temporal claims or asserting cloud availability from memory.
- `eval-10 pricing-scope` — user asks for a dollar figure on the migration; tests pricing-scope boundary (skill declines dollar figures, routes to the public cost estimator as a Sales handoff).
- `eval-11 connector-decision` — user has 12 MSK Connect connectors with 3 custom plugins; tests that the skill surfaces the managed-vs-self-managed decision and asks clarifying questions instead of auto-picking.

### What's deferred to V0.2

Four state-file-path evals (assess-iam-private, hidden-connectors, full-plan-open-questions, assess-hidden-cluster) and their shared `msk-acme-statefile/state.json` fixture were scoped out of V0.1 after early-review feedback. The original fixture was a sanitized customer scan — too large for the published skill and too risky for residual data leakage even after sanitization (`sanitize_kcp.py` scrubs identifiers but not numeric values). A 42 KB synthesized replacement was tried but it referenced the original for structural cues, leaving the same residual risk.

The clean path: regenerate from clean room (KCP schema doc only, no reference to any prior fixture) in V0.2, then re-add the four evals. The 7 manual-`migration-profile.yaml` evals already exercise equivalent scenarios (Red Flags audit, hidden-connector inference, full Plan template, cost-reconciliation logic) — the V0.2 add-back is about restoring the matched-pair design that validates input-format equivalence directly.

This is tracked in the project's V1 roadmap as the top-priority post-V0.1 task.

## Review

- [x] SME reviewer identified: @AhmedsZamzam (migration domain SME)
- [x] DTX/DevRel reviewer assigned: @Cerchie (DTX reviewer working with author on this submission); team `@confluentinc/dtx` and `@confluentinc/developer-advocates` auto-tagged via CODEOWNERS

### Architecture highlights

- **Lazy-load dispatcher.** `SKILL.md` is a router with cross-stage decision tables and non-negotiable rules; stage-specific procedures load on demand from `references/`. SKILL.md stays ~240 lines so it loads quickly without bloating context.
- **ROUTE vs HARDCODED protocol.** Stable judgment lives in the skill; volatile product facts (cluster caps, feature availability per cloud, version cutoffs) are fetched live via `WebFetch`. Hardcoded values are explicitly labeled and carry conditional framing rather than date-stamps that decay.
- **Public-sources-only.** All claims trace to docs.confluent.io, github.com/confluentinc/kcp, or general Kafka operational standards. No internal sources. No customer data.
- **Both intake paths supported.** KCP state file (when KCP is installed and AWS credentials are available) or manual `migration-profile.yaml` (walkthrough of 11 question groups). V0.1 actively tests the manual path across 7 scenarios plus a small-state-file scenario (eval-12 serverless); V0.2 restores the matched-pair coverage for the larger state-file scenarios.
- **No auto-execution.** The skill proposes commands and asks the user before running anything; this is encoded as a non-negotiable rule in SKILL.md.

### How the eval suite was developed

Each iteration ran a five-role pipeline followed by a human gate: **runners** (one per eval per config) execute the test scenario with and without the skill loaded; **graders** evaluate every assertion against actual outputs; **aggregator** rolls up per-eval results into mean, stddev, min, and a dual-gate stopping check; **proposer** diagnoses failing assertions against runner traces and drafts discrete change files; **reviewer** filters proposals into accept / reject / flag buckets using a learned-judgment context file refined from prior human feedback. After each iteration, approved changes were applied to the skill and the loop continued until the dual-gate criterion fired.

### Dual-gate stopping criterion

The loop stops only when **both** the mean pass rate AND the minimum per-eval pass rate clear 90%. This caught a real quality issue: at iteration 4 the mean was 98.7% — a naive single-gate "mean ≥ 90%" check would have stopped there with one eval at 33% silently failing. Dual-gate enforcement forced the loop to surface and fix every weak eval before declaring convergence.

### Trajectory (key milestones)

| Iter | Mean | Min eval | Notes |
|---|---|---|---|
| 1 | 87.5% | — | KCP-path suite (13 evals); below DTX gate |
| 3 | 100.0% | — | First all-pass on KCP-path suite (mean and min) |
| 5 | 97.7% | 75.0% | Suite expanded to 18 (manual-intake mirrors added); dual-gate caught a weak eval the mean would have hidden |
| 6 | 91.0% | 33.0% | Regression — runner not loading references; recovered via runner-template fix |
| 7 | 99.1% | 83.3% | Recovery |
| 9 | 99.2% | 85.7% | Row 12 manual-intake disambiguator + structural preemptive note proposed and applied |
| post-9 spot check | 100.0% | 100.0% | **Final** — fixes spot-checked against previously-failing eval; gap closed; no regressions |

Iterations 2, 4, 8 omitted from this table for brevity. Per-iteration audit trail (benchmark.json, reviewer-summary.md, change files, human-feedback notes) retained in a development workspace and available on request.

### Human-review checklist

In response to @Cerchie's [note](https://github.com/confluentinc/agent-skills/pull/25#discussion_r3275431965) on moving the checklist out of `/evals` so it reaches reviewers: the canonical source remains at [`skills/msk-migration/evals/human-review-checklist.md`](https://github.com/confluentinc/agent-skills/blob/add-msk-migration-skill-v2/skills/msk-migration/evals/human-review-checklist.md) so anyone running the eval suite locally has it adjacent to `evals.json`. The full contents are inlined below for review visibility.

The checklist covers what the [agentskills.io spec](https://agentskills.io/skill-creation/evaluating-skills) places in human review rather than LLM-graded assertions — qualities like writing style, ordering, asking-before-acting, conversational decline, adjacency / proximity in produced artifacts. It pairs with `evals.json`: `evals.json` covers substring-matchable artifact assertions; this checklist covers the rest. Both should clear before declaring an iteration done.

Two grading rules apply to every item:

1. **Items are artifact-checkable.** Every item is answerable by reading the produced artifact (Plan or Assess) and the transcript. "Would a customer understand this?" is post-hoc reasoning, not artifact grading.
2. **Items test the final correct state.** No item asks to confirm the skill avoided a prior bad framing. Negative-form items are graded in their positive form.

<details>
<summary><b>Full human-review checklist (click to expand)</b></summary>

#### Global scan: D1 operational-guidance leakage

Apply this scan to every Plan output (Evals 2, 6, 16, 19, 20, 21, 22, 23, 24, 25, 26, 27). The Plan is a Technical Plan and stays architectural — pilot validation, testing strategy, validation gates, runbook timing, and operational rollout guidance belong in the customer's operational plan, not the Plan output.

- [ ] **Plan stays architectural.** No section of the Plan contains substrings: 'pilot rollout', 'pilot validation', 'testing strategy', 'validation gates', 'runbook timing', or operational rollout language. A closing note that operational concerns are out of Technical Plan scope is acceptable and expected; specific operational guidance is not.

#### Eval 1 — kcp-fresh-start

- [ ] **Introduces KCP on first mention.** When KCP first comes up, the skill briefly explains what it is — Confluent's open-source migration tool, a link to `github.com/confluentinc/kcp`, and its purpose in the migration context. The skill does not assume the user knows what KCP is.
- [ ] **Presents the scan sequence before running anything.** The 6-step KCP scan sequence appears in the transcript before any command is invoked.
- [ ] **Does not auto-execute bash commands.** The skill does not run `kcp --version`, `kcp version`, or any bash command on its own. It either asks the user to run the command and share output, or explicitly asks whether the user wants the skill to run it.

#### Eval 2 — plan-scram-baseline

- [ ] **Cluster Type Decision recommends Enterprise with no trigger named.** The profile (1 cluster, peak egress 540 MBps, 2400 partitions) is within Enterprise eCKU caps. Cluster Type Decision section names Enterprise as the target and contains no Row 1-N trigger reference or eCKU/partition cap-exceeded claim.
- [ ] **Auth Approach target matches the profile's source auth or names an explicit basis for divergence.** Profile lists `auth_types: [scram]` only. Recommended CC target auth is SCRAM-compatible (SCRAM-SHA-512, SCRAM-SHA-256, or API key) OR the section cites a source-grounded reason for an IAM/mTLS target (customer-driven preference or named CC constraint).

#### Eval 3 — dedicated-large-fleet

- [ ] **Does not over-escalate dimensions that aren't triggered.** Recommendations do not pull along changes to auth, networking, SR, or other dimensions beyond what the profile's hard-limit triggers require. A Dedicated escalation for capacity does not produce unrelated recommendations.
- [ ] **Does not escalate without a trigger.** If another cluster in the same profile does not trigger any hard-limit row, the skill keeps it on Enterprise. Dedicated escalation is per-trigger, not a blanket recommendation.

#### Eval 4 — intake-triage

- [ ] **Does not assume an intake path.** The skill does not dive into architecture recommendations, stage execution, or environment assumptions without first establishing what the user has.
- [ ] **Routes via stage menu OR presents intake options.** The opening either (a) presents the Mode Detection stage menu (Assess / Plan / out-of-scope) so the user can signal stage before intake-path selection, OR (b) offers KCP deep scan and Manual intake as the two intake options. The skill does not jump past both into architecture, sizing, or fabricated environment claims.
- [ ] **Asks what the user has.** The skill asks either where the user is in the migration (stage signal) or which intake path the user can produce, rather than recommending a path before knowing.
- [ ] **Does not fabricate environment details.** The skill does not invent cluster configurations, auth types, region, or scale assumptions to keep the conversation moving.

#### Eval 5 — manual-no-kcp

- [ ] **Walks the intake groups in a logical order without jumping around.** The skill proceeds through platform / cluster topology / scale / auth / networking / ACLs / Schema Registry / connectors / costs / migration drivers in a coherent sequence rather than jumping around based on what the user mentioned first. Kafka version may be asked alongside platform basics (Group 1) since it's platform-adjacent.

#### Eval 8 — out-of-scope-baremetal

- [ ] **Recognizes out-of-scope source.** The skill identifies that self-managed Kafka is outside the MSK-only scope.
- [ ] **Declines to run the workflow.** The skill does not proceed with the MSK migration workflow against a non-MSK source.
- [ ] **Does not map non-MSK concepts onto MSK primitives.** The skill does not ask for ARNs, CloudWatch access, Glue SR config, or other MSK-specific artifacts as if they applied to a bare-metal source.
- [ ] **Redirect targets are limited to real, documented resources.** When the skill points the user elsewhere, the named target is one of: docs.confluent.io, the Confluent account team / Sales, or another resource explicitly documented in `SKILL.md` or its references. No sibling skill name, internal tool, or partner program is named that isn't traceable to those sources.

#### Eval 9 — mtls-azure-availability

- [ ] **Consults the feature-availability reference rather than asserting from general knowledge.** The skill routes the user to a live docs page for current feature availability, rather than asserting current availability state from training data.

#### Eval 10 — pricing-scope

- [ ] **Frames pricing as a Sales handoff, not a skill limitation.** The explanation is that pricing is handled via Sales by design, not that the skill lacks the data.
- [ ] **Does not attempt to compute cost from scan data.** The skill does not combine scan outputs with pricing guesses to produce a synthetic estimate.

#### Eval 11 — connector-decision

- [ ] **Surfaces the decision rather than picking blindly.** The skill treats managed-vs-self-managed as a decision the user needs to make, not committing to one without input.
- [ ] **Asks clarifying questions.** The skill asks whether a managed equivalent exists on Confluent Cloud for each connector type, whether source code is available for the custom plugins, and what throughput needs are.
- [ ] **Consults the connector reference.** The decision framework traces to the references rather than being improvised in the response.

#### Eval 12 — assess-serverless

- [ ] **Does not fabricate throughput values for serverless.** The mock has `peak_ingress_mbps: null` and `peak_egress_mbps: null` — serverless returns limited metrics. The output does not invent throughput numbers and does not silently pass over the missing fields.

#### Eval 14 — assess-iam-private-manual

- [ ] **Does not fabricate values for missing profile fields.** Every environment fact in the Environment Summary traces to a specific field path in the YAML profile. The skill does not invent values where data is missing.

#### Eval 15 — hidden-connectors-manual

- [ ] **Recommends closing the connector gap before Plan commits connector paths.** The skill flags that follow-up data collection is needed before the Plan stage can commit a connector migration path — either re-running `kcp scan self-managed-connectors` (if the user later runs KCP) or asking the user to enumerate the Connect fleet manually.

#### Eval 16 — full-plan-open-questions-manual

- [ ] **Sizing committed with math, not deferred.** Sizing produces concrete eCKU values per cluster with the computation visible. The skill does not defer with "TBD", "requires live fetch", or "decide at Provision". If a live fetch is needed for per-eCKU values, the skill either fetches or offers to.

#### Eval 17 — assess-serverless-manual

- [ ] **Does not fabricate throughput values for serverless (manual intake).** The mock has `peak_ingress_mbps: null` and `peak_egress_mbps: null` — serverless returns limited metrics, and the profile's `known_gaps` entry documents this. The output does not invent throughput numbers and does not silently pass over the missing fields.

#### Eval 18 — assess-hidden-cluster-manual

- [ ] **Recommends broadening discovery scope before Plan commits.** The Row 16 response recommends re-running `kcp discover` across all regions and all relevant AWS accounts (or, in manual intake, asking the user to enumerate any clusters not yet listed) before the Plan scope is committed.
- [ ] **Does not fabricate specs for the hidden cluster.** The skill does not invent broker count, topics, throughput, or other details for the unmatched cost entry. The cost signal confirms existence and gives a rough size cue (from instance type + spend), but specifics require enumeration of the hidden cluster.

#### Eval 20 — switchover-pattern-mechanism

- [ ] **Pattern × Mechanism alternatives table renders all four required rows.** Incremental + Gateway, Big-bang + Gateway, Big-bang + Manual CL, and Dual-write each have their own row. Each row's rationale column contains substantively different content — no two rows share identical or near-identical rationale text.
- [ ] **Incremental + Gateway row carries a RECOMMENDED tag on the row itself.** Tag is in column 1 or 2 (or appended to the pattern label), not a separate sentence elsewhere in the section.
- [ ] **"Incremental + Manual CL not recommended" appears as its own paragraph, callout, or labeled row note.** Standalone block — not embedded inside another row's prose or hidden in a parenthetical.
- [ ] **Dual-write row prose contains all three substrings: operational-complexity language, dual-cost language, "no Confluent-specific tooling" (or "generic CL only").** All three present in the Dual-write row.

#### Eval 21 — per-connector-classification

- [ ] **Each source connector has its own row with a distinct classification.** Fixture has s3-sink and custom-internal-validator. s3-sink row contains CMU + `kcp create-asset migrate-connectors` path. custom-internal-validator row contains the defer-to-account-team path. Two rows, two paths.
- [ ] **Substitution-nuance row, if present, names a specific substitution issue.** Substring anchor: "Debezium 1.x", "V2 config key", "config key difference", or equivalent named issue — not a generic "config differences" or "talk to Sales" phrase.

#### Eval 22 — networking-pni-default

- [ ] **PNI recommendation and cost reasoning are adjacent in the Networking Decision section.** PNI recommendation sentence is immediately followed (within the same paragraph or the next paragraph) by cost reasoning containing both "cross-AZ" anchor AND one of "data processing" / "hourly endpoint" anchors.
- [ ] **Three citable PrivateLink exceptions and the deferral cases live in separate blocks.** Three citable exceptions (cc_egress_required, gateway limit, non-AWS target) each appear in their own bullet/row with the trigger labeled. Deferral cases (compliance, organizational policy, latency-sensitivity, substrate) are grouped in a distinct bullet/row or sub-section — not interleaved with the citable exceptions.

#### Eval 23 — cl-direction-source-initiated

- [ ] **CL Direction section contains a visible derivation line.** Section contains a derivation of the form `target_networking={value} → reachability={value} → direction={value}` (or equivalent inline cascade format). Direction is derived, not configured — the cascade chain is on the page.
- [ ] **Source-initiated manual-setup callout contains all three anchors in one block.** Single paragraph/list/callout contains: (a) the manual-setup action ("establish the link manually" or "customer-owned setup"), (b) the doc citation `private-networking.html`, (c) the permanence note ("link.mode=SOURCE permanent" or "cannot be changed after the link is created"). All three within the same block — not scattered across separate paragraphs.

#### Eval 24 — schema-linking-no-outbound

- [ ] **Schema Migration section presents the selected path with cascade reasoning for the fixture's specific SR type.** EITHER (a) opens with the full 6-row source-SR-type enumeration (Confluent SR / Glue / Karapace / clean-break / adopt / skip) before the cascade, OR (b) presents the cascade-result for the fixture's specific source SR type with derivation prose that names the SR type AND names the selected path (Schema Linking via Schema Exporter, defer to account team, clean break, adopt, or skip). Per iter-19, the trimmed cascade form is acceptable when only one path applies to the fixture — showing 5 irrelevant rows is noise, not value.
- [ ] **Plan names source-side outbound reachability as the SL gating constraint.** Substring anchor present in the cascade reasoning: "source-side outbound", "source SR can reach CC SR", "outbound reachability", or `source_sr_can_push_to_cc_sr`. Anchor appears in the cascade prose, not just a passing mention elsewhere.

#### Eval 25 — auth-target-oauth

- [ ] **Three target identity models each paired with their specific CC SASL/SSL mechanism.** OAuth row/bullet names SASL/OAUTHBEARER. API Keys row/bullet names SASL/PLAIN. mTLS row/bullet names SSL with client certs. Three distinct mechanism strings, one per identity model.
- [ ] **Auth Approach has two visibly separate blocks: source-side pre-migration, then target-side identity choice.** Block 1 cites Invariant 8 and the IAM → SCRAM/mTLS source pre-migration path. Block 2 emits the OAUTHBEARER / PLAIN / SSL target cascade. Two blocks separated by a header, list boundary, or table boundary — not collapsed into a single paragraph.

#### Eval 26 — mtls-azure-matrix-fork

- [ ] **Plan names the specific matrix-row outcome inline with cluster-types.html citation.** Cluster Type Decision or Auth Approach section contains all four anchors in the same paragraph, list item, or table row: cluster-types.html URL + mTLS substring + cluster-type-name (Enterprise / Freight / Dedicated) + Azure-or-all-clouds qualifier. The matrix-row outcome must be named (e.g., "Enterprise mTLS is AWS-only", "Dedicated supports mTLS on all clouds", or equivalent that identifies which cluster-type-and-cloud combination fails the row).
- [ ] **Plan resolves the mTLS-on-Azure fork without recommending Enterprise + mTLS on Azure.** Either escalates Cluster Type Decision to Dedicated (with the mTLS row cited as the trigger) OR keeps Enterprise and routes target auth to SCRAM or API keys (with the matrix outcome named). Plan does NOT recommend the combination of Enterprise cluster type + mTLS auth + Azure target in a single recommendation block — that combination is a setup-time blocker.

#### Eval 27 — historical-data-handling-not-required

- [ ] **Tiered-storage callout names all three specifics as distinct dimensions.** Within the callout: (a) "re-fetch" + "S3" co-occur (any phrasing — "re-fetch from S3", "S3 re-fetch", "re-fetch all of it from S3" all match the data-source-from-S3 anchor); (b) "backfill time" or "time to backfill" or equivalent dimension-named substring (must be distinct from cost); (c) "backfill cost" or "cost of backfill" or "re-fetch cost". Three distinct dimensions named — conflated phrases like "material cost and time" do not satisfy because they collapse two dimensions into one.
- [ ] **not_required defer-to-account-team row contains a one-line rationale.** Row prose contains substring matching "not a documented default" or "customer-specific setup required" or equivalent rationale anchor — not just "defer to account team" with no reason given.
- [ ] **Historical Data Handling and Invariant 5 / MSK-decommissioning timing appear in distinct headed blocks.** Consumer history decision and MSK decommissioning timing live in separate sections, sub-sections, or labeled paragraphs. Not collapsed into a single bullet.

#### Eval 31 — topic-readiness-manual-opt-out

- [ ] **Opt-out note contains both data-shape anchors.** Note contains substring "manual profile" (or "profile schema") paired with "per-topic broker configs" (or "broker-side configs") AND contains either `topics.details[]` or "KCP state file" / "kcp scan clusters". Both anchors present — explaining the data-shape limitation, not a capability limitation.
- [ ] **Opt-out note enumerates at least two concrete next steps.** Step 1: re-run with KCP scan (substring anchor: "kcp scan" or "AWS credentials"). Step 2: capture configs manually for high-risk topics (substring anchor from this set: "tiered storage", "cleanup.policy", "RF" / "replication factor", or "unclean leader"). Both steps present.

</details>

### What I'd value feedback on

1. **Architecture.** Dispatcher pattern, ROUTE/HARDCODED protocol, references-load-on-demand discipline. Highest-cost decisions to change later.
2. **Eval methodology.** Does the eval ralph loop's dual-gate + proposer/reviewer pattern match what DTX wants to see for skill validation? The infrastructure is reusable and could template how other skills validate themselves.
3. **MVP scope decision.** Assess + Plan as V0.1 with V0.2 scoped to extension stages. Is "scoped MVP" the right framing, or should V1 cover all 6 stages before submission?
4. **Family pattern.** This is the first of a planned migration-skill family. `osk-migration` (open-source Kafka takeouts) is next. Should the family share a base reference structure, or stay siblings without inheritance?

Open to all other feedback.

### Author

Andy Morgan, Growth PM, Confluent. Authored with Claude Code (Opus 4.7).


leeec1993 (commented):
## Skill Review — `msk-migration`

### Overall Assessment

This is a high-quality skill with strong architecture. The ROUTE/HARDCODED protocol for handling stale knowledge, the lazy-load dispatcher pattern, and the 16-row Red Flags audit are standout features. The eval suite is rigorous on Assess-stage edge cases. There are some issues to address before merging.

---

### Must-Fix (3 issues)

**1. Frontmatter says "plan and execute" but scope is Assess + Plan only**
The description overpromises. A user triggering the skill expecting execution guidance will hit a redirect. Change to "assess and plan a migration" to match the declared MVP scope.

**2. Customer name leak in `plan.md`**
The networking section references "Coinbase" by name — needs to be genericized to "the organization" or "the user" before shipping publicly.

**3. README says 13 evals; `evals.json` has 14**
Factual error in the README. Should say 14.

---

### Should-Fix (6 issues)

**4. Plan template section count mismatch**
`SKILL.md` and README say "12-section Plan template" but `plan.md` lists 13 sections. Reconcile the count.

**5. No explicit out-of-scope handling in Mode Detection**
The Mode Detection table routes Assess and Plan but has no "Out of scope" row. The frontmatter's negative triggers are the only guard. Add an explicit decline row that echoes the negative triggers.

**6. Three orphaned mock fixtures with no evals**
`msk-provisioned-scram`, `multi-cluster-mixed`, and `partial-plan-state` are unreferenced by any eval. Either add evals for them or remove them from V0.1.

**7. `assess.md` vs eval contradiction on intake group ordering**
`assess.md` says "Ask these in whatever order fits the conversation" but eval #5 asserts "Walks the 11 intake groups in order." One of them needs to change.

**8. No `tiered_storage` or `cost_line_items` fields in `migration-profile.yaml` template**
Both are tested in evals (Row 12, Row 16) but missing from the canonical template. Eval fixtures extend the schema silently.

**9. P95 sizing from manual profiles is unspecified**
`plan.md`'s sizing procedure assumes time-series data for P95 computation. Manual profiles only carry peak values. The fallback behavior should be documented explicitly.

---

### Nice-to-Have (5 items)

**10. HARDCODED values have no refresh process**
The "last verified 2026-04-20" date on uncited HARDCODED rows will go stale with no update mechanism. Consider a maintenance note.

**11. `kcp-commands.md` is thin**
Missing `cluster-credentials.yaml` template, `kcp scan self-managed-connectors`, `kcp create-asset migrate-acls iam`, and common flags. It reads more like an index than a reference.

**12. Eval #16 has a monolithic table-format assertion**
Tests 8 cluster-table columns + 7 sizing-table columns + 3 risk-table columns in one assertion. Should be split for diagnosability.

**13. Eval #4's disjunctive acceptance is too permissive**
"Routes via stage menu OR presents intake options" accepts two fundamentally different behaviors. Should enforce stage-menu path for generic intent.

**14. No eval for multi-cluster, SCRAM auth, or populated metrics**
The suite is strong on edge cases but missing coverage for common real-world scenarios. The `multi-cluster-mixed` fixture is sitting there unused.

---

### Architecture Feedback (responding to the author's questions)

**1. Architecture** — The dispatcher pattern and ROUTE/HARDCODED protocol are excellent. The HARDCODED sub-typing (doc-cited with conditional framing vs. uncited with date stamps) is a genuinely clever solution to the stale-knowledge problem. Worth templating for other skills.

**2. Eval methodology** — The dual-gate stopping criterion (mean AND min ≥ 90%) is the right call — the example of catching a 33% eval hidden by a 98.7% mean proves its value. The five-role pipeline is thorough.

**3. MVP scope** — Assess + Plan as V0.1 is the right framing. The manual-profile path exercises the same logic as state-file path, so coverage is adequate. Ship this, then extend.

**4. Family pattern** — Recommend shared base references (auth mapping table, networking decision logic, invariants) with DB-specific overrides. The ROUTE/HARDCODED protocol should definitely be shared infrastructure, not reimplemented per skill.

---

*Reviewed by Erick Lee with Claude Code*

skills/msk-migration/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Scope (MVP)
+
+This skill covers **Assess and Plan** stages of an MSK to Confluent Cloud migration. It produces an Assessment document (Red Flags audit, Environment Summary) and a Plan document (cluster type, sizing, networking, auth, switchover approach, pre-migration workstream). It does **not** execute Provision, Migrate, Switchover, or Monitor stages — those are scoped for the next iteration. When the user asks about downstream execution (target cluster provisioning, Cluster Linking setup, client cutover, post-cutover monitoring), redirect to [docs.confluent.io](https://docs.confluent.io) and the Confluent account team rather than fabricating coverage. The Plan stage still emits structured decisions about networking, auth, switchover approach, and pre-migration steps — those decisions feed downstream execution that the user (or a future skill version) carries out.
+
+IMPORTANT: Do NOT read all reference files upfront. Read only the reference file for the stage the user is currently in. Load references lazily as needed. Exception: the decision tables, invariants, and Sources of Truth map below are always available — they are referenced across multiple stages.
```
Cerchie: This note on loading ref files is not needed as any skill that follows the spec (which yours does) will progressively load by default.

skills/msk-migration/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Cross-Stage Decision Logic
+
+### Cluster type: default to Enterprise, escalate to Dedicated only on hard limits
+
+Enterprise is the recommended target for every migration. It is elastic, supports private networking (PrivateLink and PNI on AWS), supports mTLS on AWS, supports BYOK on all clouds, supports client quotas, and covers the vast majority of migration scenarios without the operational overhead of CKU sizing. Recommend Dedicated **only** when the source environment triggers one of the hard limits below.
+
+**How to use the hard-limits table:**
+
+- **ROUTE rows require a live fetch before answering.** Do not answer an escalation question from memory. Fetch the cited doc section, extract the value, and cite the URL in the response.
```
Cerchie, line 123: Again, a bit worried about relying on routes to hard-coded links, would be interesting to see a solution that leveraged MCP as discussed offline

skills/msk-migration/evals/evals.json
```diff
@@ -0,0 +1,196 @@
+{
+  "skill_name": "msk-migration",
+  "evals": [
+    {
+      "id": 1,
+      "slug": "kcp-fresh-start",
+      "prompt": "we haven't started — ready to migrate off msk to confluent cloud. let's try kcp.",
+      "expected_output": "Skill introduces what KCP is before first use, presents the 6-step scan sequence, and asks before running any command (doesn't auto-execute bash).",
+      "files": [],
+      "assertions": [
```
Cerchie, line 10: In general, assertions should be true/false, about the resultant artifact. Otherwise there's some wiggle room for the model to assert a pass when it's really a fail. Trying to think through what an artifact of this skill would be though, as I'm not familiar with MSK migrations.

skills/msk-migration/SKILL.md
amorgan-cflt: Agreed — removed in this update.

skills/msk-migration/SKILL.md
amorgan-cflt: Continuing from the Slack thread — sticking with canonical Confluent-owned URLs as ROUTE pointers + live fetch via WebFetch + HARDCODED protocol for non-routable facts. Erick's reasoning in Slack lined up with the design: links treat docs as source of truth and avoid the dual-update problem. The skill also routes to `github.com/confluentinc/kcp` for command surface and Zero-Cut prerequisites — those aren't in the MCP allowlist, so WebFetch covers more ground than MCP could regardless. Adding a 404-fallback callout per your Slack suggestion as a small resilience improvement.

skills/msk-migration/evals/evals.json
amorgan-cflt: Did a full check of the suite against the spec (https://agentskills.io/skill-creation/evaluating-skills). What are your thoughts on these two changes to address the feedback:

**For #12 (eval #16 monolithic table-format assertion):** splitting into three separate assertions — Cluster table format, Sizing table format, Risks table format. One assertion per table.

**For the assertion-shape concern:** moving behavioral assertions to a human-review checklist per the spec's guidance — *"Not everything needs an assertion. Some qualities ... are hard to decompose into pass/fail checks. These are better caught during human review. Reserve assertions for things that can be checked objectively."* Behavioral checks like "asks before running KCP" or "walks intake groups in order" are exactly what the spec puts in human review. They move to `evals/human-review-checklist.md` (mirroring the spec's `feedback.json` per-test-case pattern), graded during manual rounds. Assertions remaining in `evals.json` stay artifact-grounded.

Does this address your feedback for V0.1?

amorgan-cflt:
@leeec1993 quick architecture question on nice-to-have #11 (kcp-commands.md is thin):

For V0.1 my plan is to keep the file as a curated index but add brief judgment-style entries for the specific items you called out (`cluster-credentials.yaml` template, `kcp scan self-managed-connectors`, `kcp create-asset migrate-acls iam`) and hoist the "route to KCP repo" guidance to the top so it's clear the index is by design, with the canonical KCP repo as source of truth for everything else. These are the architecture options I came up with:

1. **Curated index + explicit routing (my plan).** Keep judgment-style entries for common commands, route everything else to KCP repo. Lightweight; preserves judgment value-add ("0 clusters → check AWS credentials," "MSK Serverless returns limited metrics").
2. **Pure reference.** Fill out the file with comprehensive command/flag coverage. Heavier maintenance; risks staleness across KCP releases.
3. **Pure routing.** Eliminate the curated index entirely; route every command to the relevant KCP doc page. Most minimal-maintenance, consistent with the docs.confluent.io routing pattern you endorsed in Slack. Trade-off: we lose the judgment notes that are the actual value-add over plain command listings.

Which fits your read of best practice?

amorgan-cflt:
## V0.1 review feedback — Tier A changes

This push addresses 12 of the 17 review items. Two structural items are still in discussion via inline comments (Reply 3 on `evals.json`, Reply 4 on `kcp-commands.md`). One item is a V0.2 self-improvement carry-over captured in the project's V1 roadmap.

### Lucia Cerchie (DTX) — inline comments

| Comment | Change |
|---|---|
| Lazy-load directive at `SKILL.md:12` not needed (spec handles progressive disclosure by default) | Paragraph removed |
| Concern about WebFetch reliance for ROUTE-row links | Architecture unchanged (continuing the Slack-thread direction). Added a 404-fallback callout in `SKILL.md` Sources of Truth per Erick's Slack suggestion: surface 404s to the user, fall back to HARDCODED with conditional framing, flag the URL for update |

### Erick Lee — Must-Fix

| # | Feedback | Change |
|---|---|---|
| 1 | Frontmatter "plan and execute" overpromises (scope is Assess + Plan only) | Changed to "assess and plan" in `SKILL.md` frontmatter and `README.md` opening line |
| 2 | "Coinbase" customer name leak in `plan.md` | Replaced with "the user" |
| 3 | README says 13 evals; `evals.json` has 14 | Fixed both stale references — `README.md:16` ("18 scenarios" → "14") and `README.md:49` ("13 behavioral test scenarios" → "14") |

### Erick Lee — Should-Fix

| # | Feedback | Change |
|---|---|---|
| 4 | "12-section Plan template" mismatch (template is 13 sections) | Reconciled to 13. README updated in 2 places; eval #16 first assertion updated to name "Inputs & Default Assumptions" and change "12 required sections" → "13 required sections" |
| 5 | No out-of-scope row in Mode Detection | Added a third row covering Provision / Migrate / Switchover / Monitor intents — declines and redirects to docs.confluent.io and the Confluent account team |
| 7 | `assess.md` "any order" vs eval #5 "in order" contradiction | Updated `assess.md` to "Walk these in order during manual intake" — matches the eval's canonical behavior |
| 8 | `tiered_storage` and `cost_line_items` missing from `migration-profile.yaml` template | Added both. `cost_line_items` at top level (shape mirrors the `msk-hidden-cluster` fixture); `storage_mode` and `peak_tiered_storage_gb` per cluster. `assess.md` Rows 12 and 16 updated to reference the new fields so the schema is wired through end-to-end |
| 9 | P95 sizing fallback for manual profiles unspecified | Added `p95_ingress_mbps` / `p95_egress_mbps` to the manual template as preferred sizing inputs; `peak_*` retained as fallback. `plan.md` Capacity Sizing Procedure is now explicit: when only peak is provided, size on peak AND emit a prominent overestimation flag in three places (top of Sizing section, Inputs & Default Assumptions row, Open Questions) so the user knows the eCKU number is conservative-bound and can refine by providing P95 |

### Erick Lee — Nice-to-Have addressed in V0.1

| # | Feedback | Change |
|---|---|---|
| 10 | HARDCODED values have no refresh process | Added a "Maintenance for uncited HARDCODED rows" bullet in `SKILL.md` "How to use the hard-limits table": quarterly review cadence, verification via the relevant Confluent product team or internal docs, convert-to-ROUTE protocol when a structured public doc emerges |
| 11 | `kcp-commands.md` is thin | Rewrote with curated-index framing — hoisted "route to KCP repo" guidance to the top, added a "When to route to the KCP repo" section, added judgment entries for `cluster-credentials.yaml` template, `kcp scan self-managed-connectors`, and `kcp create-asset migrate-acls iam`. Architecture question posed in a separate inline reply (Reply 4) asking whether pure-reference or pure-routing would fit better |

### Awaiting reviewer direction

- **Reply 3 (assertion methodology — Lucia + Erick #12):** structural framing per the agentskills.io spec — split eval #16's monolithic table-format assertion AND move ~23 behavioral assertions to a human-review checklist. The inline reply on `evals.json` describes the proposal; not yet applied pending confirmation.
- **Reply 4 (Erick #11 architecture):** `kcp-commands.md` is implemented per Option 1 (curated index + routing). Inline reply asks which of three architectures — curated index, pure reference, or pure routing — fits best. Open to revising before merge.
- **Downstream of Reply 3:** adding two new evals using existing orphan fixtures (`msk-provisioned-scram` baseline + `multi-cluster-mixed` multi-region) — addresses Erick #6 and #14 in one move. Held until Reply 3's assertion methodology is confirmed so the new evals follow the same discipline.

### V0.2 carry-over

- Erick **#13** (eval #4 disjunctive acceptance) — root cause is Mode Detection ambiguity in the skill rather than the assertion text. Tracked in `v1-roadmap.md` for V0.2 alongside the broader Mode Detection cleanup.

skills/msk-migration/evals/evals.json
amorgan-cflt: Confirmed and applied — per your Slack ack on 2026-05-14. Pushed in commit f746720:

- 27 behavioral assertions moved from `evals.json` to a new `evals/human-review-checklist.md`. These are qualities the [agentskills.io spec](https://agentskills.io/skill-creation/evaluating-skills) puts in human review rather than LLM-graded assertions (writing style, ordering, asking-before-acting, conversational decline, etc.).
- Eval #16 monolithic table-format assertion split into 3 (Cluster, Sizing, Risks) per Erick #12.
- Remaining 83 assertions tightened to deterministic checks only — substring, regex, count, file-existence, YAML parse, specific-value match. No behavioral verbs left in `evals.json`.
- Added 2 baseline evals (`plan-scram-baseline`, `plan-multi-region-mixed-auth`) closing the orphan-fixture and multi-cluster coverage gaps from Erick #6 / #14.

Eval-loop iter-13 on the tightened suite: with-skill mean 98.04% vs no-skill baseline 23.02% (delta +75pp), 13 of 16 evals fully pass. Min eval at 87.5% — two real skill-content gaps remain (region citation in multi-region Networking Decision sections, complete `Working assumption:` label coverage across every Open Question). These are separate from the DTX feedback addressed here, and we'll continue tightening the eval suite as we apply the migration-SME feedback round.

Cerchie: Nice updates on the evals. For the human-review checklist, I'm not sure that it would reach a human in the /evals folder. Perhaps move to the PR description?

amorgan-cflt:
**Eval performance after this update**

- With skill: **99.35%** mean pass rate
- Without skill: **43.19%** mean pass rate
- 30 of 31 evals at 100%

skills/msk-migration/evals/evals.json
amorgan-cflt: I kept the file there, but lifted the human evals out for visibility. Is that okay or did you want me to remove those entirely?

Cerchie: I think ok to lift out entirely as it's less likely that a human will see it as it's invoked from an agent interface, and the agent will likely ignore as well

amorgan-cflt:
**Eval fix: removed a brittle assertion from eval-2 (`plan-scram-baseline`)** — commit `217e234`

@AhmedsZamzam flagged a failing assertion while testing the latest skill. On investigation the skill output was correct; the assertion was producing a false positive.

**What was wrong**

The eval-2 assertion `"BIDIRECTIONAL not triggered for default-case intake"` failed if the substring `recommend`/`recommended` appeared within 100 characters of `BIDIRECTIONAL` in the Switchover Approach section. That proximity heuristic cannot distinguish a plan that *excludes* BIDIRECTIONAL from one that *recommends* it. `references/plan.md` explains the exclusion in prose ("MSK is destination-initiated only... `SOURCE` and `BIDIRECTIONAL` both require a cluster-link object on the source, which open-source Apache Kafka excludes"), and that sentence naturally sits near recommendation language. Correct output, false fail.

**Why removing it is safe — coverage is preserved and better-placed**

- **eval-42** (`plan-consumers-after-producers-destination-initiated`) already asserts `"Plan does not offer BIDIRECTIONAL for the MSK source"`, on the fixture where the failure mode is actually plausible (consumers migrate after producers — the textbook case where a model might reach for bidirectional CL).
- **eval-23**, **eval-28**, and **eval-42** positively assert destination-initiated CL direction. BIDIRECTIONAL-absence is a consequence of those positive checks, not a separate fact to police.
- eval-2's SCRAM baseline carries no consumers-after-producers signal, so the assertion guarded against a mistake the fixture would never surface.

**Net:** eval-2 drops from 10 to 9 assertions; suite stays at 33 evals. Removing a false-fail can only raise the pass rate, so no dual-gate concern on re-run.

Cerchie (commented):
Thank you for the changes. Let's merge once @AhmedsZamzam approves

robertosemp (approved):
Looks good to me.

Cerchie (approved):

leeec1993 (approved):

Cerchie (approved):

PR #27, opened 2026-05-08
Add a WarpStream shared config reference, and update Skills to reference WarpStream-specific configurations

cfgrillo:
## Description

This PR adds a shared reference for WarpStream configs, which are different from the recommendations and defaults for Apache Kafka. This PR also updates several existing skills in the repo to account for these configs, and also to provide application-specific recommendations (i.e., implications of enabling EOS on Kafka Streams).

## Docs

N/A, updates existing skills

- [ ] If adding a new skill, I have reached out to reviewers to make sure the docs are updated to reflect the new skill.

## Testing

- [ ] Evals pass at 90%+ threshold

## Review

- [ ] SME reviewer identified if adding a new skill: 
- [ ] DTX/DevRel reviewer assigned

## Additional Context

WarpStream docs: docs.warpstream.com


Cerchie:
Thanks for the PR, Caleb! I will clicktest, but first, can you ensure Warpstream cases are covered by evals? And can you move the markdown folder from /shared to /references, since it is used as a reference? thanks

cfgrillo:
> Thanks for the PR, Caleb! I will clicktest, but first, can you ensure Warpstream cases are covered by evals? And can you move the markdown folder from /shared to /references, since it is used as a reference? thanks

Done @Cerchie !

Cerchie:
Got a python producer/consumer set up and running very quickly against a WS playground account with the new adds to the python client skill.

Screenshot 2026-05-11 at 9 41 05 AM

But the Python skill overall defaults to JSON though given guidance from the clients team. Initially though the WS response at `GET /schemas/types` returned ` ["AVRO","PROTOBUF"]`, so I think we need to weave this back into the guidance.

Cerchie:
PS -- clicktesting the KStreams skill went smoothly

cfgrillo:
> Got a python producer/consumer set up and running very quickly against a WS playground account with the new adds to the python client skill.
> 
> Screenshot 2026-05-11 at 9 41 05 AM
> But the Python skill overall defaults to JSON though given guidance from the clients team. Initially though the WS response at `GET /schemas/types` returned ` ["AVRO","PROTOBUF"]`, so I think we need to weave this back into the guidance.

@Cerchie  This is because you're using the WarpStream Schema Registry when you run Playground mode, which does only support Protobuf and Avro. But if the user is using any other SR, such as the Confluent Cloud SR, then JSON will work. So I think we just need to update the skill to include this decision point. Just pushed a commit that does that.

Cerchie:
still need some adjustments it would seem
```
ite-packages/confluent_kafka/schema_registry/common/__init__.py", line 27, in init
    await obj.__init__(*arg, **kwarg)
TypeError: AsyncAvroDeserializer.__init_impl() got multiple values for argument 'schema_registry_client'
```

Cerchie:
running into one more Warpstream/client incompatiblity -- thanks for your patience on this. Claude's DX: 
```
The WarpStream producer overrides set
  message.max.bytes=64000000, but those settings apply to the whole
  client (producer and consumer), and the consumer overrides only
  bump fetch.max.bytes to ~50MB. librdkafka requires fetch.max.bytes 
  >= message.max.bytes, so the consumer rejects the config.
```

Cerchie (approved):
re-tested, then tested a non-WS case to make sure that was still smooth 

 working like a charm 🚀 🚀 🚀

PR #41, opened 2026-06-05
confluent-cloud-dbt skill

fdolce:
## Description
This PR adds the confluent-cloud-dbt skill.
The skill helps LLMs avoid usual pitfalls deriving from the peculiar nature of this adapter, which fits flinks streaming concepts in dbt which is generally used for non-streaming databases.

### Eval results — `confluent-cloud-dbt`

  Ran all 6 evals in two configurations (with the skill vs. a no-skill baseline), grading every assertion against the generated files and transcripts.

  | Metric | With Skill | Baseline (no skill) | Delta |
  |--------|-----------|---------------------|-------|
  | **Pass rate** | **100%** (44/44) | 29% | **+71 pts** |
  | Time | 81.5s | 60.0s | +21.5s |
  | Tokens | 25,898 | 14,212 | +11,686 |

  ### Per-eval (with skill)

  | Eval | Scenario | Result |
  |------|----------|--------|
  | 0 | Scaffold windowed aggregation from scratch | 14/14 ✅ |
  | 1 | `incremental` not supported → route to `streaming_table` | 7/7 ✅ |
  | 2 | WHERE-clause edit → `dbt run` SKIP debug | 6/6 ✅ |
  | 3 | `dbt test` unreliable on unbounded streaming table | 6/6 ✅ |
  | 4 | `streaming_source` misuse on existing topic | 6/6 ✅ |
  | 5 | `changelog.mode='upsert'` conceptual Q&A | 5/5 ✅ |

  ### What the baseline confirms

  The no-skill runs failed in exactly the ways the skill is designed to prevent:

  - **eval-1**: baseline generated an `incremental` + `merge` model (a compile error on dbt-confluent); the skill routed correctly to `streaming_table`.
  - **eval-4**: baseline built a generic dbt `source()` staging model with no awareness of `streaming_source` semantics.
  - **eval-0/2/3**: baseline hedged on the adapter and missed Confluent-specifics (TVF `TUMBLE` form, the `--full-refresh` silent-failure mode, the 1000-row streaming-cursor cap).
  - **eval-5** was closest — baseline scored 4/5, only missing the `PRIMARY KEY ... NOT ENFORCED` metadata semantics.

  Comfortably clears the 90% eval threshold.


fdolce:
It seems CI is failing due to a link in confluent-skill-creator/SKILL.md line 172, the check seems clean for files introduced in this PR now

skills/confluent-cloud-dbt/references/adapter-behaviours.md
```diff
@@ -0,0 +1,7 @@
+# Known adapter behaviours (0.2.x bugfixes worth surfacing)
```
Cerchie: is there somewhere dynamic we can link out to? I imagine these will get fixed eventually and render the skill out of date

skills/confluent-cloud-dbt/SKILL.md
```diff
@@ -0,0 +1,121 @@
+---
+name: confluent-cloud-dbt
+description: "Authoring, scaffolding, or modifying dbt projects for the dbt-confluent adapter — Confluent Cloud's managed Apache Flink SQL service ONLY (not self-managed Flink, not other Flink runtimes). Triggers on existing dbt-confluent projects (`profile.type: confluent`, `dbt-confluent` in `pyproject.toml`/`requirements.txt`) and when the user wants to scaffold a new dbt-confluent project. Do NOT trigger for non-Confluent dbt adapters, plain Flink SQL outside dbt, or self-managed Flink clusters."
```
Cerchie: Let's add the cdc-tableflow skill to the do NOT trigger... clause just to be extra safe

skills/confluent-cloud-dbt/SKILL.md
```diff
@@ -0,0 +1,121 @@
...
+metadata:
```
Cerchie: let's remove metadata and compatibility fields just for now, we haven't standardized these at Confluent

skills/confluent-cloud-dbt/references/adapter-behaviours.md
fdolce: Right, this will get stale with time. I restructured this part a bit so that rather than having the quirks of a single specific version here, we direct claude when an error surfaces.
The idea is that when there's an error, claude first checks what version of the adapter is running, then checks if there are newer versions. If that's the case, it checks if the error was fixed in a later version, and only then the user should get the suggestion to upgrade. This should also save some context in general.

skills/confluent-cloud-dbt/SKILL.md
fdolce: Sure, fixed

skills/confluent-cloud-dbt/SKILL.md
fdolce: Ok, removed

fdolce:
Hi! Thanks for the review, I've addressed the comments, let me know if there's anything else.

skills/confluent-cloud-dbt/evals/evals.json
```diff
@@ -0,0 +1,124 @@
+{
+  "skill_name": "confluent-cloud-dbt",
+  "evals": [
+    {
+      "id": 0,
+      "prompt": "I want to start a new dbt-confluent project from scratch. I've already got a Confluent Cloud env, a Kafka cluster called 'orders-cluster', a Flink compute pool, and a service-account Flink API key. The source data is an existing Kafka topic 'orders' in the 'orders-cluster'. I want a 5-minute tumbling window of order counts and total revenue per customer.",
+      "expected_output": "Scaffolded project with dbt_project.yml (streaming_table default), profiles.yml.example, requirements.txt with a verified dbt-confluent pin, .gitignore, models/sources.yml declaring the existing orders topic, models/models.yml with column types and constraints for the new streaming_table, and models/<model>.sql containing the windowed aggregation using the TVF TUMBLE form. The warning-comment block is at the top of the streaming model. The user is told to run dbt commands themselves (not autonomous).",
+      "files": [],
+      "expectations": [
```
Cerchie: one last nit: let's use "assertions" here and elsewhere just to standardize

fdolce: Fixed

skills/confluent-cloud-dbt/assets/streaming_source.sql
```diff
@@ -0,0 +1,30 @@
...
+--
+-- For READ-ONLY references to topics that already exist in your cluster, do NOT use
+-- this materialization — declare a dbt `source` instead (see the sources.yml example).
+
+{{ config(
+    materialized='streaming_source',
+    connector='faker',
+    with={
+      'rows-per-second': '5',
+      'number-of-rows': '1000',     -- bound the source so dbt tests are deterministic
```
jlrobins: The comment attempt within the jinja makes a syntax error. Jinja expressions don't support -- (or #) as an inline comment token — only block-level {# ... #} is valid. Move the comment above the {{ config(...) }} block (as a plain -- SQL comment), or drop it.

I believe this is this is the only occurrence.

jlrobins (commented):
Good stuff. The new skill needs a call-out in the README.md.

I'm going to install this and use it off and on and we'll be able to revise as we learn more.

skills/confluent-cloud-dbt/assets/dbt_project.yml
```diff
@@ -0,0 +1,40 @@
...
+config-version: 2
+
+profile: my_project
+
+model-paths:    ["models"]
+seed-paths:     ["seeds"]
+test-paths:     ["tests"]
+analysis-paths: ["analyses"]
+macro-paths:    ["macros"]
+snapshot-paths: ["snapshots"]    # snapshots are NOT supported by dbt-confluent — keep this empty
```
jlrobins: "keep this empty" vs giving sample nonempty is a contradiction? Or needs comment rephrasing to differentiate between _the array being empty_ vs _the directory in the project_ being empty?

skills/confluent-cloud-dbt/SKILL.md
```diff
@@ -0,0 +1,116 @@
...
+6. **Using double quotes for identifiers.** Flink uses **backticks**. Always.
+7. **Putting `NOT NULL` / `VIRTUAL` / `METADATA` in a column's `data_type` field in `models.yml`.** A custom validator raises a clear error. Constraints belong in the `constraints:` section.
+8. **Suggesting `ALTER TABLE …` to fix anything.** All forms (rename, add/drop column, truncate) are unsupported from dbt's side. The answer is `--full-refresh`.
+9. **Telling the user to just `dbt run` to redeploy a streaming pipeline after editing the SELECT.** Without `--full-refresh`, the materialization SKIPs (column names/types unchanged → no drift detected). See the schema-drift callout below — this is a silent failure.
+10. **Treating `database` and `schema` as warehouse-style.** `database` = Confluent Environment (`env-xxxxxx`); `schema` = Kafka cluster. The adapter cannot create or drop either — both are managed in Confluent Cloud.
+11. **Suggesting `dbt seed` for non-trivial CSVs.** Single-batch INSERT, errors past the batch size. Direct users to a `streaming_source` (Datagen) or an external producer.
```
jlrobins: Perhaps should suggest the Flink `'connector'='faker'` (instead of 'Datagen') for consistency with other places?

skills/confluent-cloud-dbt/assets/dbt_project.yml
fdolce: Yep, both work, but I'll just make the default an empty list

skills/confluent-cloud-dbt/assets/streaming_source.sql
fdolce: Fixed

skills/confluent-cloud-dbt/SKILL.md
fdolce: Yep, fixed

PR #38, opened 2026-06-01
Add kafkajs-migration skill

jmathai-confluent:
## Summary

- Adds a new skill to migrate Node.js Kafka applications from KafkaJS to Confluent's `@confluentinc/kafka-javascript` client
- Covers import changes, `kafkaJS` config wrapping, producer/consumer/admin API differences, error handling, authentication, transactions, and performance tuning
- Includes 11 evals (7 positive, 4 negative) — **97% pass rate**

## Migration results (kafkajs-sample repo)

The skill was tested against a sample KafkaJS application ([confluentinc/kafkajs-sample](https://github.com/confluentinc/kafkajs-sample)):

- **Without skill**: [PR #1](https://github.com/confluentinc/kafkajs-sample/pull/1) — Claude Code without the migration skill. Took 15m 52s, tests failed due to missed migration patterns.
- **With skill**: [PR #2](https://github.com/confluentinc/kafkajs-sample/pull/2) — Claude Code with the migration skill loaded. Took 3m 0s, all tests passed.

## Skill structure

```
skills/kafkajs-migration/
├── SKILL.md                              # Workflow + quick reference
├── evals/
│   └── evals.json                        # 11 evals (7 positive, 4 negative)
└── references/
    ├── docs-migration.md                 # Official Confluent migration guide
    └── performance-guidance.md           # Production tuning (batching, memory, K8s)
```

## Validation

- `skill-validator`: 0 errors, 0 warnings
- `check_eval_schema.py`: 0 findings
- `check_trigger_overlap.py`: nits only (single keyword overlaps with unrelated skills)
- Evals: 32/33 assertions passed (97.0%)

## Test plan

- [x] Skill validator passes (0 errors, 0 warnings)
- [x] Eval schema validates (0 findings)
- [x] Trigger overlap check — no blocking overlaps
- [x] Evals pass at 97% (above 90% threshold)
- [x] Tested against real KafkaJS app with local Kafka cluster
- [x] Negative evals confirm skill does NOT trigger for: greenfield projects, node-rdkafka migrations, post-migration infra issues, existing Confluent client users

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #48, opened 2026-06-17
Kafka Java client skill

davetroiano:
## Description

Add skill for developing Kafka client apps in Java.

## Docs

- [ ] If adding a new skill, I have reached out to reviewers to make sure the docs are updated to reflect the new skill.

## Testing

## Review

- [x] `/confluent-skill-reviewer` has been run
- [] SME reviewer identified if adding a new skill
- [] DTX/DevRel reviewer assigned


skills/developing-kafka-java-client/references/share-consumer.md
```diff
@@ -0,0 +1,91 @@
+# Share Consumer API ("Queues for Kafka")
```
lianetm, line 1: Do we have a similar doc for the regular consumer?

I expect it would be valuable to provide high level context on the 2 consumer protocols available now (classic vs consumer), set via the group.protocol config etc (there's an ongoing effort to push adoption for the new protocol btw, drastically better rebalance experience in consumer groups)

skills/developing-kafka-java-client/references/share-consumer.md
```diff
@@ -0,0 +1,91 @@
...
+
+Choose the Share Consumer API when the user wants:
+- to scale consumers independently of partition count,
+- queue / work-queue semantics (competing consumers, cooperative concurrent processing),
+- per-record acknowledgement with redelivery of un-acked records.
+
+Choose the regular Consumer API (the default) when the user needs per-key ordering, exactly the
+classic consumer-group model, or has not asked for queue semantics.
+
+## Acknowledgement modes
```
lianetm, line 32: similar, consumer has the whole story of automatic commit vs manual commit, sync and async..just raising in case it's something we want to add an md similar to this but for the regular consumer

lianetm (commented):
Thanks @davetroiano !

skills/developing-kafka-java-client/references/AvroConsumer.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.time.Duration;
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * Consumer that deserializes via Schema Registry into the generated SpecificRecord
+ * {@code Transaction}. Uses the wakeup pattern for graceful shutdown: a shutdown hook
+ * calls consumer.wakeup(), the poll loop catches the resulting WakeupException, and
+ * close() runs in finally to commit offsets and leave the group cleanly.
```
lianetm, line 21: not sure if it helps to clarify that this bit of committing offset is happening only because the consumer auto commits by default (enable.auto.commit config true), so this holds while that config remains unset but not if someone builds on top and changes etc.

skills/developing-kafka-java-client/references/AppTest.java
```diff
@@ -0,0 +1,145 @@
+package com.example.kafka;
```
jmathai-confluent, line 1: I have a question about ownership of code samples in skills. Normally the docs team or eng team is responsible for fixing bugs or updating code samples. What's the ownership model for skills?

Is it an option to link to the relevant docs with this content?

cc @d8joseph

skills/developing-kafka-java-client/references/warpstream-optimization.md
```diff
@@ -0,0 +1,282 @@
...
+metadata.recovery.strategy=rebootstrap
+```
+
+**EOS note for Kafka Streams:** If the user's existing app uses `processing.guarantee=exactly_once_v2`, inform them of the tradeoff:
+
+> EOS enables idempotent producers internally (max 5 in-flight requests). On WarpStream, this reduces throughput and may produce `KAFKA_STORAGE_ERROR` retries. Consider whether `at_least_once` with downstream deduplication would work for your use case. If EOS is required, it will work — plan for additional capacity.
+
+---
+
+## librdkafka / confluent-kafka-python Overrides
```
jmathai-confluent: I think these should be removed from this skill. We will have language specific skills instead.

skills/developing-kafka-java-client/references/warpstream-optimization.md
```diff
@@ -0,0 +1,282 @@
+# WarpStream Client Optimization
```
jmathai-confluent, line 1: I think this might have been derived from the Warpstream docs - should this be a skill on its own?

skills/developing-kafka-java-client/references/share-consumer.md
davetroiano: good idea, I added this

skills/developing-kafka-java-client/references/AvroConsumer.java
davetroiano: I updated the examples to manually commit

skills/developing-kafka-java-client/references/warpstream-optimization.md
davetroiano: 👍  I removed the non-Java warpstream guidance.

skills/developing-kafka-java-client/references/warpstream-optimization.md
davetroiano: We decided when first adding warpstream guidance that we'd overlay guidance "inline" per skill though we acknowledged we need a better way to generalize. Separate skill _may_ be the way forward but it's just one option that would probably have some tradeoffs. As one example downside, I think we should allow for the possibility that users will be surgical in terms of which skills they choose to install rather than accepting the kitchen sink. so it'll need some tinkering and tradeoff weighing. for now in this PR I think we should stick with this pattern and then I can take on revisiting the general problem and how we want to handle it now that it's getting more obviously problematic when it comes to warpstream overlay guidance.

skills/developing-kafka-java-client/references/AppTest.java
davetroiano: The general model and direction that I think this is heading in is that product / eng both contribute and maintain, with the `agent-skills` repo maintainers doing the requisite notifications and reviews etc. This one is a little different because I'm contributing so I would expect to continue to be involved not in a "this is my baby" reason but a "I added this so it's only fair I have a hand in maintaining it until there's agreement to 'take it over'". probably not the best answer but it's kinda where we're at given that the maintainers want to get to that cleaner ownership model but at the same time we are bootstrapping the repo so there are in between cases like this.

davetroiano: regarding public docs, we do link to docs in the repo. I tend to be a bit hesitant since docs weren't necessarily written with agent skills in mind which might have some quality / control issues when it comes to skill effectiveness, token usage, and potential for a rug pull with a docs change that isn't keeping their spot as a dependency in mind. There are some other things that I think need some figuring out. e.g., docs publishes markdown files to be agent friendly but if those markdown files contain relative links elsewhere on Confluent Docs, they'll be html. over time we might get more comfortable with the idea of skills pointing out to canonical references more often but the comfort isn't there today at least speaking for myself. even today if I look at the docs page I'd naturally point to for this, it'd be `https://docs.confluent.io/kafka-clients/java/current/overview.md` which contains html links as well as links over to Confluent Developer examples. it's really human-intended.

skills/developing-kafka-java-client/references/share-consumer.md
davetroiano: 👍  I added coverage for these topics in the consumer reference

Cerchie (changes requested):
I ran to create both docker and CC apps, and when I created an app for CC I ran into a version issue, as pom.xml pinned org.apache.avro:avro to 1.11.4 as a direct dependency, and Confluent's kafka-avro-serializer:8.2.1 is compiled against Avro 1.12.1

skills/developing-kafka-java-client/SKILL.md
```diff
@@ -0,0 +1,303 @@
+---
+name: developing-kafka-java-client
```
Cerchie, line 2: Let's add metadata.version as we had a recent update for tracking versions here

skills/developing-kafka-java-client/references/warpstream-optimization.md
jmathai-confluent: That sounds okay to me. Tx.

skills/developing-kafka-java-client/references/AppTest.java
jmathai-confluent: Agree. In fact, I did similar on a PR of my own which @Cerchie [suggested linking out](https://github.com/confluentinc/agent-skills/pull/38#issuecomment-4614033070).

Just flagging. Feel free to ignore.

Cerchie (approved):
worked smoothly this time around!

PR #50, opened 2026-07-09
skill creator and skill reviewer: update skill wording + script to check for PII 

Cerchie:
Adds a script to check for the presence of PII, and adds that script and verbiage to check for PII to the skill reviewer and skill creator. Also adds it to CI.


davetroiano (approved):
lgtm!

leeec1993:
@Cerchie should we have an allowlist for test card numbers?  
everything else looks good from my end

Cerchie:
> Lucia Cerchie (@Cerchie) should we have an allowlist for test card numbers? everything else looks good from my end

good thought, I can address in a follow-on

PR #28, opened 2026-05-12
 initial skill-reviewer

Cerchie:
## Description

Adds a skill-reviewer to the Confluent skills and top-level .claude folder.

PR #19, opened 2026-04-17
DTX-991 - fix python client 

Cerchie:
## Description

This addresses[ DTX-991 ](https://confluentinc.atlassian.net/browse/DTX-991) by a) adding examples for producing with keys to the references snippets and b) adding a mandatory gate in the SKILL md for confirming user choices 

## Skill Quality Checklist

- [x] Skill [description is optimized](https://agentskills.io/skill-creation/optimizing-descriptions) for accurate triggering
- [x] Feedback loops like unit tests are built-in to enable agent iteration

## Testing

- [x] Evals pass at 90%+ threshold

## Review

- [ ] SME reviewer identified if adding a new skill: 
- [x] DTX/DevRel reviewer assigned

PR #54, opened 2026-07-30
Add confluent-cloud-flink-sql skill

Cerchie:
## Description

Ports [ksilin/cc-flink-sql-skill](https://github.com/ksilin/cc-flink-sql-skill) into this repo as `confluent-cloud-flink-sql`, enforcing the Confluent Cloud vs. Apache Flink OSS dialect boundary (CC-vs-OSS traps, `confluent` CLI flag schemas, CC-validated SQL patterns, supported serialization formats, CC-specific troubleshooting, reserved words).

Converted the source repo's eval harness (JSONL trap prompts + Python scorer) into this repo's `evals/evals.json` format, added the skill to the README table, and bumped the plugin version (1.2.1 → 1.3.0) per the versioning convention in CLAUDE.md.

## Test plan

- [x] SME reviewer assigned for Flink SQL / Confluent Cloud domain @ksilin 
- [x] DTX/DevRel reviewer assigned @davetroiano 
- [x] Run the skill's evals and confirm ≥90% pass rate, paste score in this PR: │ Total       │ 41/41 (100%)
- [x] Spot-check a few `references/*.md` files render correctly and links resolve on GitHub

🤖 Generated with [Claude Code](https://claude.com/claude-code)


eguicf (changes requested):
Thank you for the PR! I've tested a few times, with `/confluent-skill-reviewer` and without. Here are  some issues worth address: 

```
Blocking (3)

1. skills/confluent-cloud-flink-sql/references/cli-reference.md:7 — --cloud/--region are not rejected on statement create.
The skill asserts this in four places: cli-reference.md:7,14 ("❌ rejected"), SKILL.md:54 (red flag), troubleshooting-cc.md:90-95 ("Error: unknown flag: --cloud"), and evals.json:101 (assertion). The official CLI reference lists both flags in the Cloud flag set for confluent flink statement create:

--environment string  Environment ID.
--cloud string        Specify the cloud provider as "aws", "azure", or "gcp".
--region string       Cloud region for Flink ...

That same page marks only --sql as REQUIRED. --environment is not required (the doc's own example omits it entirely), while cli-reference.md:7 claims "✅ required". This is most likely CLI version drift: some older version really did reject the flag. The problem is that it is already encoded in eval 10's assertion, so per CLAUDE.md § Evals are the contract it will be locked in as a regression test for a wrong fact. Re-verify the whole flag schema table against the pinned CLI version, then fix all four locations plus the eval assertion.

2. skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md:217 — the MATCH_RECOGNIZE DEFINE clause uses LAG(), which is not valid on CC.

N AS N.temperature_c > 90 AND N.temperature_c > LAG(N.temperature_c),

CC's MATCH_RECOGNIZE offers only logical offsets, LAST(variable.field, n) and FIRST(variable.field, n), for navigating within a pattern variable. LAG is an OVER window function and cannot appear inside DEFINE. The correct form is (LAST(N.temperature_c, 1) IS NULL OR N.temperature_c > LAST(N.temperature_c, 1)). This file is titled "Verified for Confluent Cloud", and eval 20's assertions only forbid PREV(/NEXT(, so the invalid form scores a clean pass. Fix the pattern and add an assertion to eval 20 requiring the LAST(...) logical offset.

3. skills/confluent-cloud-flink-sql/SKILL.md:56 — CRUD-capable skill with no plan-before-execute step.
Verification loop step 2 goes straight to confluent flink statement create, which creates a real statement, consumes CFUs, and for DDL also creates topics. cli-reference.md:108 hands the agent confluent kafka topic delete my-topic --cluster $CLUSTER --force, and cli-reference.md:122 references subject delete --permanent. Grepping the whole skill finds no confirmation language at all (the only hit is the comment # Wait for deletion propagation). Phase B rule 7 of this reviewer requires an explicit "present the plan, wait for user confirmation" step before any resource-modifying call. Add it after EXPLAIN (step 0) and before statement create, and gate the delete snippets the same way.

Warning (5)

- references/reserved-words.md:7 — the must-backquote list is inaccurate. Against CC's Flink SQL Reserved Keywords page: payload is not in the keyword list at all, and name, key, type, data are nonreserved, usable unquoted. The rest of the list (timestamp, value, time, offset, partition, start, end, and so on) does check out as reserved. Either drop those five, or rename the section to something like "column names that commonly need backquoting" and state plainly that it is an empirical list, not the reserved set.
- references/cli-reference.md:17 — "--property is comma-separated in ONE flag ... (NOT repeated flags)". The CLI docs type it as --property strings, a slice flag that accepts both comma-separated values and repeated flags. The first half is right; "NOT repeated flags" has no basis and should be deleted.
- SKILL.md:4 — frontmatter carries only metadata.version, missing metadata.author: confluent and metadata.last_updated. Both other in-repo skills (confluent-skill-creator, confluent-skill-reviewer) declare them; this is the CLAUDE.md provenance convention.
- SKILL.md:84 — - "Don't invent identifiers — use <placeholder> for anything you haven't verified."  sits inside the Anti-patterns list, where every other bullet is a thing not to do. Read literally, it tells the agent that avoiding invented identifiers is an anti-pattern, which inverts the intent. It is also the only quoted bullet and has a trailing space. Move it to Non-negotiables and drop the quotes.
- SKILL.md:20,61,64 — workflow assumptions carried over from the source repo (ksilin/cc-flink-sql-skill) that write files into the user's own project: non-negotiable 3 mandates docs/*.md and docs/flink-dialect-traps.md; the verification loop mandates repro/<phase>-<slug>.sql and docs/VERIFICATION-<phase>.md. The Source-of-truth hierarchy at SKILL.md:96-98 additionally tells the agent to upstream newly found traps back into this skill, meaning it edits the installed skill files. For external users these are all unrequested writes. Make them optional suggestions, or require user consent first.

Nit (3)

- No compatibility: field, despite clear environment requirements (the confluent CLI, an available CC compute pool that will burn CFUs, optionally Terraform). Add one modeled on confluent-skill-creator.
- Coverage: the skill maps OSS-to-CC traps thoroughly but says nothing about CC-only DDL that has no OSS counterpart (CREATE MATERIALIZED TABLE, CREATE MODEL / AI_COMPLETE, CREATE AGENT, USE CATALOG). The description promises "Write and debug Flink SQL that runs on Confluent Cloud", so a user asking about those gets nothing. If the scope is deliberately narrowed to the dialect boundary, say so explicitly in SKILL.md.
- Eval coverage: none of the 20 evals exercises reserved-word backquoting or the --property form, so references/reserved-words.md has no eval protection at all.
```
Correction on Blocking 2 from another agent test: EXPLAIN against a real CC compute pool (AWS us-west-2) plans LAG(N.temperature_c) inside MATCH_RECOGNIZE DEFINE successfully, so the example is valid and might be unblocking.

eguicf:
Thanks for the quick turnaround @Cerchie ! Most items are addressed. A few new things, plus one correction on me for `LAG()`. The warning and nit part below is worth looking into too.

```
Blocking (3)

1. A stale copy of this skill is committed at .claude/skills/confluent-cloud-flink-sql/

Error: fdf02de added 8 files / 1128 lines of the pre-review draft, and .gitignore has no .claude entry. It carries the same name:, so anyone working in this repo loads the pre-fix draft as a project-level skill: no plan-before-execute confirmation, and the false claim that --cloud/--region are rejected on statement create. 6 of the 8 files already diverge from skills/.
Fix:
git rm -r .claude/skills/
echo '.claude/skills/' >> .gitignore

2. Four flink artifact snippets fail as written

Error: --cloud and --region are marked REQUIRED on artifact create, list, and describe.
$ confluent flink artifact list
Error: required flag(s) "cloud", "region" not set

Missing at SKILL.md:75, cli-reference.md:44, cli-reference.md:45, troubleshooting-cc.md:119. (cli-reference.md:43 already has them, and troubleshooting-cc.md:141 is fine.)

Fix: Add --cloud <provider> --region <region> to those four lines. Then close the gap that caused it: the flag table at cli-reference.md:7-13 covers only flink statement, so the artifact rules are undocumented. Add a note after it:
- flink artifact create/list/describe: --cloud and --region are REQUIRED, unlike flink statement where both are optional.
- flink compute-pool: no --cloud flag exists. list accepts --region; describe accepts only --environment.

If you extend the table instead, do not keep --cl. They diverge for compute-pool.

3. A false claim about LAG()/LEAD() inside DEFINE, enforced by an eval

Error: LAG() inside MATCH_RECOGNIZE DEFINE plans successfully on CC (EXPLAIN reaches COMPLETED on a real compute pool, two separate runs). So the "not valid there" clause at sql-patterns-cc.md:201 is wrong, and evals.json:202 asserts Does NOT use LAG( or LEAD( inside DEFINE, which means a correct response using LAG() scores as a failure and the clause cannot be removed without the eval flagging it.
Fix: Delete both. The claim appears nowhere else in the references, and the LAG(...) OVER (...) section at sql-patterns-cc.md:299 is unaffected. Keep the assertion requiring LAST(...)/FIRST(...), but reword its reason to "documented pattern-navigation construct, handles the first-row NULL case." Only planning was verified, not what LAG evaluates to inside MATCH_RECOGNIZE, so avoid asserting a specific se

Warning (2)

4. dialect-traps.md:42-46 contradicts SKILL.md:10
Error: The reference says "Hit a new trap, add row here," while SKILL.md:103 forbids editing the skill's own installed files. References load standalone, so the reference wins.
Fix: Reframe the section as maintainer guidance, or drop it.

5. Eval 8 passes a state-losing answer
Error: The prompt never establishes whether the statement is stateful, and neither assertion requires mentioning the stateless-only limit documented at cli-reference.md:93 (Stateless statements only (no aggregates/windows/pattern
matching/upsert sinks)), so an answer that omits
Fix: Add an assertion requiring the answer to sta to stateless statements only.

Nit (4)

6. dialect-traps.md:45 names a nonexistent skill,: confluent-cloud-flink-sql.
7. cli-reference.md:85 uses jq -r '.status // .stThis works today because the CLI returns a flat"status", but // short-circuits on a truthy object, so the .status.phase fallback is dead code and the parser is not actually defensive as the note above it implies. .status | if type == "object" then .phase else . end handles both shapes.
8. SKILL.md:45-59 restates 13 of the 32 traps. Cond place to keep in sync.
9. None of the three skills named in this skill's confluent-cloud-cdc-tableflow,kafka-streams-programming) reciprocate. Overlap is only on the broad token flink, so cosmetic.
```

Cerchie:
removed the extraneous .claude changes

### `Error: --cloud and --region are marked REQUIRED on artifact create, list, and describe.`
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/SKILL.md#L75
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/references/cli-reference.md#L48-L49
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/references/troubleshooting-cc.md#L119
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/references/cli-reference.md#L16-L17

### `A false claim about LAG()/LEAD() inside DEFINE, enforced by an eval`
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md#L201
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/evals/evals.json#L203

### `dialect-traps.md:42-46 contradicts SKILL.md:10`
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L40-L42

### `Eval 8 passes a state-losing answer`
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/evals/evals.json#L83

### `cli-reference.md:85 uses jq -r '.status `
https://github.com/confluentinc/agent-skills/blob/f80a30529512eb25eaa418e66356e3298ba2666b/skills/confluent-cloud-flink-sql/references/cli-reference.md#L89

eguicf (approved):
LGTM! A few non-blocking suggestion and we can always improve in follow-ups: 

1. `SKILL.md:43` — The "Red flags" list restates 13 of the 32 rows from
   references/dialect-traps.md, including the fixes. An agent can act on it
   without ever loading the reference, and the two lists will drift apart.
   Keep only the trigger signal here (construct name plus "see trap N") and
   leave the fix in the reference.

2. `references/sql-patterns-cc.md:317` — The example writes `$rowtime as
   event_time`, which is exactly what line 322 warns against. The warning is
   correctly scoped to CTEs, but the code block is what gets copied, and trap 8
   rates aliasing CRITICAL. Add an inline comment in the snippet or drop the
   alias.

3. `references/sql-patterns-cc.md:262,266` — Late Data Routing splits rows on
   `>= CURRENT_WATERMARK(...)` and `< CURRENT_WATERMARK(...)`. Before the first  watermark fires CURRENT_WATERMARK() is NULL, both predicates are NULL, and  those rows go to neither sink. A routing pattern should not silently drop data. Add an IS NULL branch or document the gap.

4. The `.gitignore` rule for `.claude/skills/` is not in the PR.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+|---|---|---|---|---|
+| 1 | `CREATE CATALOG ...` | Catalog = env, database = cluster; not creatable | HIGH | CC UDF project |
+| 2 | `SET 'execution.checkpointing.interval'` | CC manages checkpointing; not user-settable | HIGH | CC UDF project |
+| 3 | `CREATE TABLE ... WITH ('connector' = 'kafka', ...)` | Tables auto-mapped from Kafka topics; manual DDL limited | HIGH | CC Flink project |
+| 4 | `WITH cte AS (...) INSERT INTO ...` | `INSERT INTO ... WITH cte AS (...)` — CTE comes AFTER INSERT INTO (empirical) | HIGH | CC Flink project |
+| 5 | `CREATE TEMPORARY VIEW v AS (...)` | `TEMPORARY` not supported — use persistent `CREATE VIEW` or CTE (`WITH v AS (...)`) | MEDIUM | CC Flink project |
+| 6 | `INSERT INTO ... VALUES` | Materializes; test use only, not production | LOW | CC UDF project |
+| 7 | `t.$rowtime` (qualified system column) | Custom event-time columns with explicit watermarks ARE supported. For auto-mapped tables, use `$rowtime`; never qualify as `t.$rowtime` | MEDIUM | CC Flink project |
+| 8 | `$rowtime AS alias` in CTE | Silently strips time-attribute property — downstream watermark breaks (empirical) | CRITICAL | CC Flink project |
```
MartijnVisser: **[blocker] Trap #8 is refuted by a live CC run.** Aliasing `$rowtime AS event_time` in a CTE and windowing on it with `TUMBLE(TABLE t, DESCRIPTOR(event_time), ...)` works — the time-attribute property is preserved, so this CRITICAL "silent strip" is wrong. Remove/rewrite, and update eval #6 which depends on it.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+
+| # | Apache Flink SQL (OSS) | Confluent Cloud Flink SQL | Severity | Source |
+|---|---|---|---|---|
+| 1 | `CREATE CATALOG ...` | Catalog = env, database = cluster; not creatable | HIGH | CC UDF project |
+| 2 | `SET 'execution.checkpointing.interval'` | CC manages checkpointing; not user-settable | HIGH | CC UDF project |
+| 3 | `CREATE TABLE ... WITH ('connector' = 'kafka', ...)` | Tables auto-mapped from Kafka topics; manual DDL limited | HIGH | CC Flink project |
+| 4 | `WITH cte AS (...) INSERT INTO ...` | `INSERT INTO ... WITH cte AS (...)` — CTE comes AFTER INSERT INTO (empirical) | HIGH | CC Flink project |
+| 5 | `CREATE TEMPORARY VIEW v AS (...)` | `TEMPORARY` not supported — use persistent `CREATE VIEW` or CTE (`WITH v AS (...)`) | MEDIUM | CC Flink project |
+| 6 | `INSERT INTO ... VALUES` | Materializes; test use only, not production | LOW | CC UDF project |
+| 7 | `t.$rowtime` (qualified system column) | Custom event-time columns with explicit watermarks ARE supported. For auto-mapped tables, use `$rowtime`; never qualify as `t.$rowtime` | MEDIUM | CC Flink project |
```
MartijnVisser: **[blocker] Trap #7 is refuted.** `SELECT o.$rowtime FROM ... AS o` (qualified with the table alias) runs fine on CC. Drop "never qualify as `t.$rowtime`".

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+| 21 | `debezium-json` (OSS naming) | CC uses `json-debezium-registry` (format-first naming) | MEDIUM | CC Flink project |
+| 22 | Savepoints, `STOP WITH SAVEPOINT` | Not exposed on CC; statement deletion = state loss; use `prevent_destroy` in TF | CRITICAL | CC Flink project |
+| 23 | Savepoint-based statement upgrade | NOT available on CC — savepoints internal-only; cannot transfer state to different statement | CRITICAL | CC Flink project |
+| 24 | DataStream API | Not supported. Table API (Java/Python) IS supported; DataStream is not | HIGH | CC docs |
+| 25 | `LATERAL TABLE(UNNEST(...))` | Parse error on CC. Use `CROSS JOIN UNNEST(...)` | HIGH | CC Flink project |
+| 26 | `PROCTIME()` function | Not supported on CC. Use External Tables/KEY_SEARCH_AGG, upsert-kafka join, or event-time temporal join | HIGH | CC docs / community |
+| 27 | `CURRENT_TIMESTAMP` in update-producing queries | Rejected as non-deterministic | MEDIUM | CC docs / community |
+| 28 | `SET 'sql.state-ttl' = '...'` as separate statement | Not allowed. Pass via `--property sql.state-ttl=ms` | MEDIUM | CLI experience |
```
MartijnVisser: **[blocker] Trap #28 is wrong.** `sql.state-ttl` is a supported `SET` option (see statements/set.md, default 0 ms). `--property` is an extra path, not the only one. Fix here and in eval #19.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+| 18 | `GROUP BY TUMBLE(ts, INTERVAL '1' MINUTE)` | `TUMBLE(TABLE t, DESCRIPTOR(ts), INTERVAL '1' MINUTE)` (table-valued functions required) | HIGH | CC UDF project |
+| 19 | `CREATE FUNCTION f AS '...'` | Must include `USING JAR 'confluent-artifact://<id>'` + `USING CONNECTIONS` for egress | HIGH | CC UDF project |
+| 20 | `'value.format' = 'json'` | NOT supported. Use `'json-registry'` (or `avro-registry`, `proto-registry`, `raw`). CC supports 7 formats only | HIGH | CC Flink project |
+| 21 | `debezium-json` (OSS naming) | CC uses `json-debezium-registry` (format-first naming) | MEDIUM | CC Flink project |
+| 22 | Savepoints, `STOP WITH SAVEPOINT` | Not exposed on CC; statement deletion = state loss; use `prevent_destroy` in TF | CRITICAL | CC Flink project |
+| 23 | Savepoint-based statement upgrade | NOT available on CC — savepoints internal-only; cannot transfer state to different statement | CRITICAL | CC Flink project |
+| 24 | DataStream API | Not supported. Table API (Java/Python) IS supported; DataStream is not | HIGH | CC docs |
```
MartijnVisser: **[blocker] Reword trap #24.** DataStream unsupported is right, but "Table API (Java/Python) IS supported" overstates it. PTF (the ProcessFunction replacement, SQL-invocable) + Table API are GA in Java; Python Table API is Open Preview with no PTF yet. See flink/concepts/process-table-functions.html.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+| 25 | `LATERAL TABLE(UNNEST(...))` | Parse error on CC. Use `CROSS JOIN UNNEST(...)` | HIGH | CC Flink project |
```
MartijnVisser: **[fix] Trap #25 is mischaracterized.** `LATERAL TABLE(UNNEST(...))` isn't valid Flink syntax anywhere (not just CC), and plain `LATERAL (subquery)` works fine on CC. Reword to "use CROSS JOIN UNNEST for arrays"; don't imply LATERAL is unsupported. Also fix eval #4's "parses on OSS but fails on CC" premise.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+Single source of truth. Update here; per-project CLAUDE.md references this file.
+
+## Trap table
+
+| # | Apache Flink SQL (OSS) | Confluent Cloud Flink SQL | Severity | Source |
+|---|---|---|---|---|
+| 1 | `CREATE CATALOG ...` | Catalog = env, database = cluster; not creatable | HIGH | CC UDF project |
+| 2 | `SET 'execution.checkpointing.interval'` | CC manages checkpointing; not user-settable | HIGH | CC UDF project |
+| 3 | `CREATE TABLE ... WITH ('connector' = 'kafka', ...)` | Tables auto-mapped from Kafka topics; manual DDL limited | HIGH | CC Flink project |
+| 4 | `WITH cte AS (...) INSERT INTO ...` | `INSERT INTO ... WITH cte AS (...)` — CTE comes AFTER INSERT INTO (empirical) | HIGH | CC Flink project |
```
MartijnVisser: **[structural] Split "CC-vs-OSS" from "general Flink".** #4, #11, #12, #13, #14, #16, #27 (and likely #6; #15 is obsolete under TVF-only) are accurate but identical in OSS — not dialect differences. Add a "CC-specific?" column or split the table, and recount the "32". Also please re-validate every remaining "(empirical)"-sourced row — two of them (#7, #8) already failed live testing.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+| 5 | `CREATE TEMPORARY VIEW v AS (...)` | `TEMPORARY` not supported — use persistent `CREATE VIEW` or CTE (`WITH v AS (...)`) | MEDIUM | CC Flink project |
+| 6 | `INSERT INTO ... VALUES` | Materializes; test use only, not production | LOW | CC UDF project |
+| 7 | `t.$rowtime` (qualified system column) | Custom event-time columns with explicit watermarks ARE supported. For auto-mapped tables, use `$rowtime`; never qualify as `t.$rowtime` | MEDIUM | CC Flink project |
+| 8 | `$rowtime AS alias` in CTE | Silently strips time-attribute property — downstream watermark breaks (empirical) | CRITICAL | CC Flink project |
+| 9 | `ALTER TABLE t ADD WATERMARK ...` | Uses `MODIFY` not `ADD` for watermarks | MEDIUM | CC Flink project |
```
MartijnVisser: **[enhance] Trap #9 is too narrow.** ALTER TABLE also does `MODIFY (`key` STRING, `val` STRING)` for schemaless/built-in-serializer topics and `SET ('value.format.schema-context'=...)`, plus adding metadata columns — not just watermarks.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+| 13 | `CAST('2024-01-01T00:00:00Z' AS TIMESTAMP)` | ISO-8601 fails — use `TO_TIMESTAMP(REPLACE(...))` | MEDIUM | CC Flink project |
+| 14 | `UNIX_TIMESTAMP(ts_column)` on TIMESTAMP | Only accepts STRING argument — cast first | MEDIUM | CC Flink project |
+| 15 | `SESSION_START(ts)` with one arg | Requires two arguments | LOW | CC Flink project |
+| 16 | `current_watermark()` on updating table | Not supported — non-deterministic function with update messages error | HIGH | CC docs |
+| 17 | UDF `open(RuntimeContext)` with FS/socket access | CC sandbox — no filesystem, no sockets without `CONNECTION` + `USING CONNECTIONS` | HIGH | CC UDF project |
+| 18 | `GROUP BY TUMBLE(ts, INTERVAL '1' MINUTE)` | `TUMBLE(TABLE t, DESCRIPTOR(ts), INTERVAL '1' MINUTE)` (table-valued functions required) | HIGH | CC UDF project |
+| 19 | `CREATE FUNCTION f AS '...'` | Must include `USING JAR 'confluent-artifact://<id>'` + `USING CONNECTIONS` for egress | HIGH | CC UDF project |
+| 20 | `'value.format' = 'json'` | NOT supported. Use `'json-registry'` (or `avro-registry`, `proto-registry`, `raw`). CC supports 7 formats only | HIGH | CC Flink project |
```
MartijnVisser: **[enhance] Make the format map precise (#20/#21).** avro-registry↔OSS flink-avro/flink-sql-avro; json-registry↔OSS json; proto-registry↔OSS flink-protobuf (OSS has ~no pure-SQL protobuf). CC also supports Debezium in JSON/Protobuf, unlike OSS.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,47 @@
...
+| 24 | DataStream API | Not supported. Table API (Java/Python) IS supported; DataStream is not | HIGH | CC docs |
+| 25 | `LATERAL TABLE(UNNEST(...))` | Parse error on CC. Use `CROSS JOIN UNNEST(...)` | HIGH | CC Flink project |
+| 26 | `PROCTIME()` function | Not supported on CC. Use External Tables/KEY_SEARCH_AGG, upsert-kafka join, or event-time temporal join | HIGH | CC docs / community |
+| 27 | `CURRENT_TIMESTAMP` in update-producing queries | Rejected as non-deterministic | MEDIUM | CC docs / community |
+| 28 | `SET 'sql.state-ttl' = '...'` as separate statement | Not allowed. Pass via `--property sql.state-ttl=ms` | MEDIUM | CLI experience |
+| 29 | `--sql-file` flag on `flink statement create` | Does not exist. Read file, pass via `--sql "$(cat file.sql)"` | HIGH | CLI experience |
+| 30 | `CREATE DATABASE` | Not supported on CC | MEDIUM | CC docs / community |
+| 31 | Aggregate UDFs (UDAF), table aggregate functions | UDAF and table aggregates not supported. Scalar UDFs and table functions (UDTF, Java only) ARE supported | HIGH | CC docs |
+| 32 | `CREATE TEMPORARY FUNCTION` | Not supported on CC | MEDIUM | CC docs / community |
```
MartijnVisser: **[blocker/add] Missing CRITICAL trap.** `DROP TABLE` deletes the physical Kafka topic + data, and `CREATE TABLE` creates the physical topic + SR schemas — unlike OSS (pure metadata). Also worth adding: topic↔table auto-mapping (default `key`/`val` BINARY) and the SELECT-vs-MATERIALIZED-TABLE materialization guidance.

skills/confluent-cloud-flink-sql/references/formats-and-serialization.md
```diff
@@ -0,0 +1,79 @@
...
+) WITH (
+  'value.format' = 'json-registry',
+  'value.json-registry.id-encoding' = 'header',   -- schema ID in Kafka header
+  'changelog.mode' = 'append'
+);
+```
+
+| `id-encoding` | Behavior |
+|---|---|
+| `confluent` (default) | Magic byte + 4-byte schema ID prepended to payload |
```
MartijnVisser: **[blocker] Wrong default value.** id-encoding values are `payload` (default) and `header`, not `confluent`. Option keys are `key.<format>.id-encoding` / `value.<format>.id-encoding`.

skills/confluent-cloud-flink-sql/references/formats-and-serialization.md
```diff
@@ -0,0 +1,79 @@
...
+Flink auto-maps the topic to a typed table. Use `SELECT src.field` instead of `JSON_VALUE(CAST(val AS STRING), '$.field')`.
+
+Caveat: registered schema types must match actual JSON types exactly. For polymorphic fields, omit from schema and use `JSON_VALUE` selectively.
+
+## Consumer output format flags
+
+| Topic encoding | `--value-format` flag |
+|---|---|
+| Raw bytes (no SR) | `string` (default) |
+| JSON Schema (SR-backed, `id-encoding=confluent`) | `jsonschema` |
```
MartijnVisser: **[blocker] Same fix.** `id-encoding=confluent` → `id-encoding=payload` (also line 62). Payload/magic-byte is the default; `header` puts the schema id in a Kafka header.

skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Temporal join (event-time, versioned table)
+```sql
+SELECT o.order_id, o.amount, o.currency, r.rate, o.amount * r.rate as amount_usd
+FROM orders o
+JOIN currency_rates FOR SYSTEM_TIME AS OF o.order_time AS r
+ON o.currency = r.currency;
+```
+
+### Regular join against upsert-kafka (lookup alternative)
```
MartijnVisser: **[blocker] Remove/strong-warn this pattern.** A regular LEFT JOIN against an upsert-kafka topic is a streaming join over an updating table — it retains the whole table in state and is inefficient. Prefer the lookup join (`KEY_SEARCH_AGG`) or an event-time temporal join (both append-only). Note CC infers upsert mode from compaction anyway.

skills/confluent-cloud-flink-sql/references/troubleshooting-cc.md
```diff
@@ -0,0 +1,164 @@
...
+## Function Errors
+
+### "Function 'PROCTIME' is not supported"
+```
+Function 'PROCTIME' is not supported in Confluent's Flink SQL dialect.
+```
+**Cause:** CC has no `PROCTIME()`. Lookup join pattern doesn't compile.
+**Fix (pick one):**
+1. External Tables + `KEY_SEARCH_AGG` (canonical for DB/REST lookups)
+2. Regular join against upsert-kafka topic (reference data in Kafka)
```
MartijnVisser: **[blocker] Same issue in the PROCTIME fix.** Drop "Regular join against upsert-kafka topic" as option 2; keep External Tables/`KEY_SEARCH_AGG` and the event-time temporal join (append-only, efficient).

skills/confluent-cloud-flink-sql/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+| [references/sql-patterns-cc.md](references/sql-patterns-cc.md) | When writing SQL — CC-validated patterns: windows, joins, dedup, MATCH_RECOGNIZE, JSON, External Tables |
+| [references/formats-and-serialization.md](references/formats-and-serialization.md) | When configuring table formats — 7 supported formats, id-encoding, consume flags |
+| [references/troubleshooting-cc.md](references/troubleshooting-cc.md) | When debugging errors — CC-specific error to cause to fix |
+| [references/reserved-words.md](references/reserved-words.md) | When hitting parse errors — must-backquote identifiers |
+
+## Red flags
+
+Stop and consult `references/dialect-traps.md` if you catch yourself writing any of these:
+
+- DataStream API (Java/Scala) — not supported on CC. Table API (Java/Python) IS supported
```
MartijnVisser: **[blocker] Reword this red flag.** "Table API (Java/Python) IS supported" overstates Python. Java Table API + PTF are GA; Python is Open Preview (no PTF). PTF is the DataStream/ProcessFunction replacement.

skills/confluent-cloud-flink-sql/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+5. **EXPLAIN before CREATE.** Always `EXPLAIN` a query before `statement create` — catches parse/type errors without consuming CFUs.
+6. **Don't invent identifiers.** Use `<placeholder>` for any topic, table, statement, or resource name you haven't verified.
+
+## Reference files
+
+Load these on demand when the topic matches — do not read them all upfront:
+
+| File | When to load |
+|------|-------------|
+| [references/dialect-traps.md](references/dialect-traps.md) | Before writing ANY Flink SQL — 32 CC-vs-OSS traps, single source of truth |
```
MartijnVisser: **[structural] "32 CC-vs-OSS traps" needs recounting** once OSS-identical rows are recategorized (see dialect-traps split). Same figure appears in the description and README.

skills/confluent-cloud-flink-sql/evals/evals.json
```diff
@@ -0,0 +1,229 @@
...
+        "Includes a `CREATE CONNECTION` step for the Postgres endpoint when using the External Tables approach"
+      ]
+    },
+    {
+      "id": 6,
+      "prompt": "Write a CC Flink SQL query that uses a CTE to prepare data, then does a tumbling window aggregation. The source table is sensor_data. Use the system rowtime column for event time.",
+      "expected_output": "A query that references $rowtime inside the CTE without aliasing it, preserving the time-attribute property needed for the downstream window.",
+      "files": [],
+      "assertions": [
+        "References the system column as `$rowtime` without aliasing it inside the CTE (not 'AS $rowtime' or 'as $rowtime'), since aliasing silently strips the time-attribute property",
```
MartijnVisser: **[blocker] Eval #6 rests on the refuted trap #8.** Aliasing `$rowtime` does not strip the time attribute (verified on CC). Rework this case/assertion.

skills/confluent-cloud-flink-sql/evals/evals.json
```diff
@@ -0,0 +1,229 @@
...
+        "Statement contains `INSERT INTO output_topic`"
+      ]
+    },
+    {
+      "id": 4,
+      "prompt": "Write CC Flink SQL to explode a JSON array field from a Kafka topic. The input table raw_events has a VARBINARY val column containing JSON with a $.items array. Extract each item as a separate row.",
+      "expected_output": "A query using CROSS JOIN UNNEST to explode the array, avoiding the LATERAL TABLE(UNNEST(...)) form that parses on OSS but fails on CC.",
+      "files": [],
+      "assertions": [
+        "Uses CROSS JOIN UNNEST(...) to explode the array, NOT LATERAL TABLE(UNNEST(...)) (parse error on CC)",
```
MartijnVisser: **[fix] Eval #4 premise is false.** `LATERAL TABLE(UNNEST(...))` doesn't "parse on OSS" — it's invalid Flink syntax generally. Keep the CROSS JOIN UNNEST expectation but drop the OSS claim (line 38 too).

skills/confluent-cloud-flink-sql/evals/evals.json
```diff
@@ -0,0 +1,229 @@
...
+      ]
+    },
+    {
+      "id": 19,
+      "prompt": "How do I set state TTL for a Flink SQL statement on Confluent Cloud?",
+      "expected_output": "Guidance to pass state TTL via the CLI --property flag on statement create, not a SQL SET statement.",
+      "files": [],
+      "assertions": [
+        "Sets state TTL via the CLI --property sql.state-ttl=<ms> flag on flink statement create",
+        "Does NOT use a SET 'sql.state-ttl' or SET 'table.exec.state.ttl' SQL statement"
```
MartijnVisser: **[blocker] Eval #19 assertion is wrong.** "Does NOT use SET 'sql.state-ttl'" contradicts the docs — `SET 'sql.state-ttl'` is valid. Accept either SET or `--property`.

MartijnVisser (changes requested):
**Overall: request changes — but this is a genuinely valuable skill.** The CC-specific core is accurate and will save real time: topic↔table auto-mapping, TVF-only windowing, format names, the PROCTIME / DataStream / savepoints / CREATE CATALOG / CREATE DATABASE restrictions, the UDF artifact+sandbox model, the CLI flag schema, External Tables + `KEY_SEARCH_AGG`, and the reserved-words list all check out. It can't merge as-is, though: a few entries are factually wrong (two verified wrong by running them on a live CC compute pool), one eval encodes a false premise, and the "32 CC-vs-OSS dialect traps" framing is inflated because ~9 entries are ordinary Apache Flink behaviors, not CC-vs-OSS differences.

_Reviewed by a Flink PMC / CC-for-Apache-Flink SME via claim-by-claim interview, live reproducers on a real compute pool, and cross-checks against the CC comparison, serialization, CLI command-reference, and statements/overview + statements/set docs._

### Must-fix (blockers)
1. **Trap #8 (CRITICAL) is wrong.** Aliasing `$rowtime AS x` in a CTE does NOT strip the time-attribute property. Verified on CC: `WITH t AS (SELECT $rowtime AS event_time, * FROM examples.marketplace.orders) SELECT window_start,... FROM TABLE(TUMBLE(TABLE t, DESCRIPTOR(event_time), INTERVAL '1' MINUTE)) ...` runs fine. Remove/rewrite, and fix eval #6.
2. **Trap #7 is wrong.** `SELECT o.$rowtime FROM ... AS o` (qualified) works on CC (verified). Drop the "never qualify as `t.$rowtime`" clause.
3. **Trap #28 + eval #19 are wrong.** `sql.state-ttl` IS settable via `SET` (statements/set.md lists it, default 0 ms). `--property` is an additional path, not the only one.
4. **id-encoding default is mislabeled.** The value is `payload` (default) or `header`, not `confluent` — fix the id-encoding table, the consumer row, and examples.
5. **Drop the "regular LEFT JOIN against upsert-kafka" lookup advice.** A streaming join against an updating table retains the entire table in state (inefficient); lookup joins and event-time temporal joins are append-only and efficient. CC also infers upsert/retract mode from the topic (compaction→upsert, Debezium schema→retract), so hand-rolling upsert-kafka is doubly off.
6. **Add the biggest CC-vs-OSS gotcha, currently missing (CRITICAL): `DROP TABLE` deletes the physical Kafka topic and its data** (and `CREATE TABLE` creates the physical topic + SR schemas) — unlike OSS where CREATE/DROP are pure metadata.
7. **Reword trap #24.** DataStream API unsupported is correct; the replacement for DataStream/ProcessFunction users is a ProcessTableFunction (PTF) in the Table API, invocable from SQL. Table API + PTF are GA in **Java**; the Python Table API is **Open Preview** and has no PTF yet — so "Table API (Java/Python) IS supported" overstates Python.

### Recategorize — the "dialect" premise is inflated
These are accurate about CC but IDENTICAL in Apache Flink OSS, so they aren't CC-vs-OSS traps: #4, #11, #12, #13, #14, #16, #27 (and likely #6; #15 is also obsolete under TVF-only windowing). #25 is mischaracterized — `LATERAL TABLE(UNNEST())` isn't valid Flink anywhere and plain `LATERAL (subquery)` works on CC, so reword to "use CROSS JOIN UNNEST" without implying LATERAL is unsupported (and fix eval #4's "parses on OSS" premise). Please add a "CC-specific vs general-Flink" column (or split the table) and recount the "32".

### High-value additions (SME)
The auto-mapping model (default `key`/`val` BINARY, replaced by real types on a matching TopicNameStrategy schema) and its one-time ALTER TABLE recipes (RecordNameStrategy/TopicRecordNameStrategy subject names; schemaless built-in serializer → `ALTER TABLE t MODIFY (`key` STRING, `val` STRING)`; schema context via `ALTER TABLE t SET ('value.format.schema-context'=...)`) — which also show trap #9's "ALTER only changes watermark" is too narrow; changelog-mode inference; the precise format-name map (avro-registry↔flink-avro, json-registry↔json, proto-registry↔flink-protobuf, ~no pure-SQL protobuf in OSS); the real SET option set (link statements/set.md; `execution.*`/`table.exec.*` rejected) plus `SET 'sql.snapshot.mode'='now'` ≙ OSS `execution.runtime-mode='batch'`; materialization semantics (SELECT is exploratory-only — 1 CFU, no recovery, streaming never ends; MATERIALIZED TABLE is the recommended long-running path; VALUES = bounded → COMPLETED); and cheap statement adds `SHOW JOBS` / `SHOW CREATE TABLE` / `DROP VIEW`.

### Nits
Commit says "20 eval cases" but there are 22; the "41/41" score in the description is stale (now 48 assertions / 22 cases) — re-run and update after fixing evals #4/#6/#19. (The `assertions[]` key is correct and matches every other skill — CLAUDE.md's `expectations[]` mention is itself stale.)

Nice work overall — once the verified-wrong entries, the id-encoding/join fixes, and the DROP TABLE trap land and the table is honestly split, this becomes a strong, trustworthy skill.

Cerchie:
1. Trap #8: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L55
2. Trap #7 (row 5): https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L17
Trap #7 (refuted note): https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L54
3. Trap #28: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L29
4. id-encoding: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/formats-and-serialization.md#L35
5. LEFT JOIN (sql-patterns-cc.md): https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md#L100
LEFT JOIN (dialect-traps.md row 16): https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L28
6. DROP TABLE: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L34
7. Trap #24: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L27
8. General Flink SQL behavior section: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/references/dialect-traps.md#L36
9. SKILL.md frontmatter: https://github.com/confluentinc/agent-skills/blob/ca7cbfd989d0a63c14b465758553a4f2d01ed895/skills/confluent-cloud-flink-sql/SKILL.md#L6-L8

skills/confluent-cloud-flink-sql/references/dialect-traps.md
Cerchie: chose the split

Cerchie:
Thanks @MartijnVisser , I've left a summary of the changes Claude made with links out to each file.

MartijnVisser (commented):
Thorough follow-up review (high effort). Six findings below as inline comments. Findings on the `id-encoding=confluent` value, the `vin` correlation column, and the plugin version sync are the ones I'd resolve before merge; the dead reference, broken code span, and eval-count mismatch are quicker fixes.

Note: one finding is about `.claude-plugin/plugin.json` (unchanged in this PR, so not inline-able there); I've attached it to the `.cursor-plugin/plugin.json` version bump instead.

skills/confluent-cloud-flink-sql/references/troubleshooting-cc.md
```diff
@@ -0,0 +1,165 @@
...
+```bash
+confluent flink compute-pool describe $POOL_ID
+confluent flink statement list --cloud <provider> --region <region>
+# Stop unused statements or increase max-cfu
+```
+
+## Consume Errors
+
+### UTF-8 decode errors on topic consume
+**Cause:** Default `--value-format string` on SR-backed topic with `id-encoding=confluent` — first 5 bytes are wire-format prefix.
```
MartijnVisser: **Nonexistent config value.** This references `id-encoding=confluent`, but that value does not exist. The only two documented values in `formats-and-serialization.md` are `payload` and `header`. For this same UTF-8 decode scenario, `formats-and-serialization.md:64` correctly uses `id-encoding=payload`. As written, an agent following this entry would report a config value to the user that is not real. Suggest changing `confluent` to `payload` to match.

skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md
```diff
@@ -0,0 +1,313 @@
...
+    fleet_id STRING
+) WITH (
+    'connector' = 'confluent-jdbc',    -- hyphen in TABLE
+    'confluent-jdbc.connection' = 'fleet_db_conn',
+    'confluent-jdbc.table-name' = 'vehicles'
+);
+
+-- Step 3: KEY_SEARCH_AGG + CROSS JOIN UNNEST
+SELECT s.signal_id, v.make, v.model, v.fleet_id
+FROM telemetry_signals s,
```
MartijnVisser, line 131: **Correlates on a column the table is never shown to have.** This canonical CC lookup-join example correlates on a bare `vin` column, but `telemetry_signals` is never given a `vin` column. Every other example in this file uses `vehicle_id` for that table (e.g. the tumbling-window example above). Copied verbatim, the query either fails to resolve the correlated column or silently correlates on the wrong field. Recommend switching to `vehicle_id` and qualifying the correlation explicitly (`s.vehicle_id`) so the intent is unambiguous.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,63 @@
...
+| 15 | DataStream API | Not supported on CC. The replacement for DataStream/`ProcessFunction` is a ProcessTableFunction (PTF) in the Table API, invocable from SQL. Table API + PTF are GA in **Java**; the Python Table API is **Open Preview** and has no PTF yet — don't claim Python parity | HIGH | CC docs |
+| 16 | `PROCTIME()` function | Not supported on CC. Use External Tables + `KEY_SEARCH_AGG` (canonical) or an event-time temporal join (`FOR SYSTEM_TIME AS OF`). Do **not** emulate a lookup join with a regular join against an upsert-kafka topic — it keeps the whole reference table in state and is inefficient; CC also infers upsert/retract mode from the topic itself (compaction → upsert, Debezium schema → retract) | HIGH | CC docs / community |
+| 17 | `SET 'sql.state-ttl' = '<ms>'` | Also supported on CC as a `SET` statement (see `statements/set.md`; default 0 ms = disabled). `--property sql.state-ttl=<ms>` on `flink statement create` is an *additional* path — useful for setting the value before the statement starts — not the only one | LOW | CLI experience |
```
MartijnVisser: **Dead reference.** This row points to `statements/set.md` for the `sql.state-ttl` detail, but no such file ships in this skill. The references added here are only `cli-reference.md`, `dialect-traps.md`, `formats-and-serialization.md`, `reserved-words.md`, `sql-patterns-cc.md`, and `troubleshooting-cc.md`. An agent following this pointer has nowhere to go. Either add the file or inline the SET-statement detail here.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,63 @@
...
+| 2 | `SET 'execution.checkpointing.interval'` | CC manages checkpointing; not user-settable | HIGH | CC UDF project |
+| 3 | `CREATE TABLE ... WITH ('connector' = 'kafka', ...)` | Tables auto-mapped from Kafka topics; manual DDL limited. Auto-mapped columns default to `key`/`val` BINARY unless a schema is registered under a matching TopicNameStrategy subject | HIGH | CC Flink project |
+| 4 | `CREATE TEMPORARY VIEW v AS (...)` | `TEMPORARY` not supported — use persistent `CREATE VIEW` or a CTE (`WITH v AS (...)`) | MEDIUM | CC Flink project |
+| 5 | `t.$rowtime` (qualified system column) | CC exposes `$rowtime` as a built-in event-time system column on auto-mapped tables. It may be referenced unqualified or qualified with a table alias (e.g. `o.$rowtime`) — both work (verified live) | LOW | CC Flink project |
+| 6 | `ALTER TABLE t ADD WATERMARK ...` | Uses `MODIFY`, not `ADD`, for watermarks — and also for other one-time schema recipes: overriding key/value types on schemaless topics (`ALTER TABLE t MODIFY (`key` STRING, `val` STRING)`), setting schema context (`ALTER TABLE t SET ('value.format.schema-context' = ...)`), and adding metadata columns | MEDIUM | CC Flink project |
```
MartijnVisser: **Broken Markdown code span.** The nested backticks around `key` and `val` inside this code span will prematurely close and reopen the span on GitHub, rendering `key` and `val` as plain-text fragments and risking column-alignment breakage for the row. Escape the inner backticks or restructure the cell so the `ALTER TABLE t MODIFY (...)` example renders as one clean code span.

skills/confluent-cloud-flink-sql/evals/evals.json
```diff
@@ -0,0 +1,147 @@
+{
```
MartijnVisser, line 1: **Eval count vs. test plan mismatch.** The PR description claims a 41/41 (100%) eval pass rate, but this merged `evals.json` contains only 14 cases. The repo's 90%+ eval gate appears to have been validated against the source repo's original 41-prompt harness, not the 14-case file actually landing here. Can you confirm the 14-case file was run through the harness and share its actual pass rate, so the merged artifact's number is demonstrated rather than inherited?

.cursor-plugin/plugin.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "confluent-streaming-skills",
-  "version": "1.2.1",
+  "version": "1.3.0",
```
MartijnVisser, line 3: **Plugin version out of sync.** CLAUDE.md requires `.claude-plugin/plugin.json` and `.cursor-plugin/plugin.json` to be bumped together and kept in sync. This PR bumps `.cursor-plugin/plugin.json` to 1.3.0 but leaves `.claude-plugin/plugin.json` at 1.2.1, so the Claude manifest reports a stale version for the same repo state after this lands. Please bump `.claude-plugin/plugin.json` to 1.3.0 as well.

skills/confluent-cloud-flink-sql/references/troubleshooting-cc.md
Cerchie: fixed here https://github.com/confluentinc/agent-skills/pull/54/changes#diff-68ca8ff82596d68e46d1b9d2f3f6e2576a2d14e86389be9435688c95e4d8446aR150

.cursor-plugin/plugin.json
Cerchie: fixed here https://github.com/confluentinc/agent-skills/pull/54/changes#diff-037fda53525563fb89230704794a06db1150204ca28ce80697a637121c3ed3f9R4

skills/confluent-cloud-flink-sql/evals/evals.json
Cerchie: yes this is fine, just a PR desc out of date

skills/confluent-cloud-flink-sql/references/dialect-traps.md
Cerchie: fixed here https://github.com/confluentinc/agent-skills/pull/54/changes#diff-e3ee2007b7ee82c9dc4d34859ac6be5af4758e84e3373a97c6ef37bc8d6b5a96R29

skills/confluent-cloud-flink-sql/references/dialect-traps.md
Cerchie: fixed, see preview

skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md
Cerchie: changed DESCRIPTOR(vin), vin to DESCRIPTOR(vin), s.vehicle_id

MartijnVisser (commented):
Second-pass review of the updated PR. The earlier round's issues (id-encoding value, dead statements/set.md reference, broken code span, plugin-version sync) all look resolved, thanks. Four findings remain below: two I'd resolve before merge and two nits. I empirically validated the two runtime claims via live CC `EXPLAIN` (compile/plan only, no jobs) on runtime flink-2.0.2553.0-v1, each with a negative control. Details inline.

One earlier concern I want to explicitly withdraw: I previously flagged the single-statement KEY_SEARCH_AGG + CROSS JOIN UNNEST(search_results) example as possibly unbindable. Live EXPLAIN COMPLETED for it, and the control with a bogus column name failed with 'Column not found', so search_results binds correctly in the combined form. That example is fine as written.

skills/confluent-cloud-flink-sql/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+
+Stop and consult `references/dialect-traps.md` if you catch yourself writing any of these:
+
+- DataStream API (Java/Scala) — not supported on CC. Table API + PTF are GA in Java; Python Table API is Open Preview with no PTF yet
+- `CREATE CATALOG ...` — catalog = CC environment, not creatable
+- `SET 'execution.checkpointing.*'` — CC-managed, not settable
+- `CREATE TABLE ... WITH ('connector' = 'kafka', ...)` — tables auto-map from topics
+- `'value.format' = 'json'` — must be `'json-registry'` (or another SR-backed format)
+- `WITH cte AS (...) INSERT INTO ...` — CC requires the CTE AFTER `INSERT INTO`
+- `$rowtime AS alias` in a CTE — silently strips the time-attribute property
```
MartijnVisser: **Contradicts this skill's own source of truth, and is empirically wrong.** This red-flag bullet says aliasing `$rowtime AS alias` in a CTE silently strips the time-attribute property. But `references/dialect-traps.md` (this skill's declared single source of truth) lists that exact claim under **"Removed traps (refuted by live testing, do not re-add)"** and states the time attribute survives the alias.

I confirmed dialect-traps.md is correct on a live CC compute pool (runtime flink-2.0.2553.0-v1):

- `EXPLAIN WITH t AS (SELECT $rowtime AS event_time, * FROM src) SELECT window_start, COUNT(*) FROM TABLE(TUMBLE(TABLE t, DESCRIPTOR(event_time), INTERVAL '1' MINUTE)) GROUP BY window_start, window_end;` -> **COMPLETED** (planner accepted the aliased column as a time attribute).
- Negative control with `DESCRIPTOR(amount)` (a non-time column) -> **FAILED**: "the timecol must be a time attribute type in streaming mode."

Since the control proves the planner does reject non-time-attribute descriptors, the COMPLETED result on the aliased column means the property survives. Please delete this bullet so SKILL.md stops contradicting dialect-traps.md.

skills/confluent-cloud-flink-sql/references/cli-reference.md
```diff
@@ -0,0 +1,148 @@
+# Confluent Cloud Flink — CLI Reference
+
+## Flag schema
+
+All `flink statement` subcommands accept `--cloud`/`--region`, `--environment`, and `--context` — none of them are rejected on any subcommand. `--sql` is the only flag marked REQUIRED anywhere in this group (on `create`); everything else, including `--environment`, is optional per the CLI's own docs (the CLI falls back to whatever `confluent environment use`/`confluent kafka cluster use` context is active).
```
MartijnVisser: **Understates required flags for `flink statement create` on CC.** This says `--sql` is "the only flag marked REQUIRED anywhere in this group" and that `--environment` is optional. Per the official CLI reference and the local CLI (v4.73.0) `--help`, `confluent flink statement create` on Confluent Cloud requires **`--sql`, `--compute-pool`, and `--database`** (and `--environment` is REQUIRED in the CLI help as well). An agent trusting this as the authoritative flag reference could omit `--compute-pool`/`--database` and either fail to schedule or run against whichever pool/cluster is active in context instead of the intended one. Please update the sentence and the table to mark `--compute-pool` and `--database` as required on `create`.

Ref: https://docs.confluent.io/confluent-cli/current/command-reference/flink/statement/confluent_flink_statement_create.html

skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md
```diff
@@ -0,0 +1,313 @@
...
+    RANGE BETWEEN INTERVAL '1' HOUR PRECEDING AND CURRENT ROW
+  ) as events_last_hour
+FROM events;
+```
+
+## LAG/LEAD Window Functions
+
+```sql
+SELECT product_id, event_time, price,
+  LAG(price) OVER (PARTITION BY product_id ORDER BY event_time) AS prev_price,
```
MartijnVisser: **Nit (efficiency):** this example evaluates the identical `LAG(price) OVER (PARTITION BY product_id ORDER BY event_time)` window three times in one SELECT (here and lines 296-297). On a high-volume stream that is the same OVER window computed three times per row for no semantic gain. Consider computing it once in a subquery (`SELECT *, LAG(price) OVER (...) AS prev_price FROM price_updates`) and referencing `prev_price` in the outer CASE. Not blocking.

skills/confluent-cloud-flink-sql/references/sql-patterns-cc.md
```diff
@@ -0,0 +1,313 @@
...
+  JSON_VALUE(CAST(val AS STRING), '$.customer.name') as customer_name,
+  JSON_VALUE(CAST(val AS STRING), '$.items[0].product_id') as first_product,
+  JSON_QUERY(CAST(val AS STRING), '$.items') as all_items
+FROM raw_topic;
+
+-- Explode JSON array (CC uses CROSS JOIN UNNEST, NOT LATERAL TABLE(UNNEST(...)))
+SELECT order_id, signal_str
+FROM input_topic
+CROSS JOIN UNNEST(
+  CAST(JSON_QUERY(CAST(val AS STRING), 'lax $.signals[*]'
```
MartijnVisser: **Nit (cleanup):** the outer `CAST(... AS ARRAY<STRING>)` wraps a `JSON_QUERY(... RETURNING ARRAY<STRING>)` that already establishes the same type, so the cast is redundant. In a canonical copy-paste pattern it obscures what `RETURNING` already does and leaves a reader unsure which type declaration matters. Consider dropping the outer CAST. Not blocking.

skills/confluent-cloud-flink-sql/references/cli-reference.md
```diff
@@ -0,0 +1,148 @@
...
+- Same org + environment + region for v1 and v2
+- v1 must be STOPPED, not deleted
+- Do NOT use `--wait` on carry-over creates
+
+## Eventually-consistent state
+
+After delete, `list` still shows resource for ~5-15s. Gate on side-effects, not existence:
+
+```bash
+# Gate on topic existence (DDL creates topics)
```
MartijnVisser, line 107: **The example contradicts the rule it is illustrating.** Line 104 states the rule: after a delete, `list` still shows the resource for ~5-15s, so "Gate on side-effects, not existence." This code block then does exactly the opposite, and its comment says so out loud: `# Gate on topic existence`, gating on a `grep` of `kafka topic list`.

That is the specific failure the rule exists to prevent. An agent that just deleted a topic and follows this sanctioned example can hit the stale-visibility window, see the topic "still exists", and skip the DDL, so the topic never gets recreated. Since this is the illustrative example directly under the rule, it is the version an agent is most likely to copy.

Suggest replacing it with a side-effect gate (for example, poll until the topic is actually absent before recreating, or make the DDL idempotent with `CREATE TABLE IF NOT EXISTS`), so the example demonstrates the rule instead of undercutting it.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,63 @@
...
+| 3 | `CREATE TABLE ... WITH ('connector' = 'kafka', ...)` | Tables auto-mapped from Kafka topics; manual DDL limited. Auto-mapped columns default to `key`/`val` BINARY unless a schema is registered under a matching TopicNameStrategy subject | HIGH | CC Flink project |
+| 4 | `CREATE TEMPORARY VIEW v AS (...)` | `TEMPORARY` not supported — use persistent `CREATE VIEW` or a CTE (`WITH v AS (...)`) | MEDIUM | CC Flink project |
+| 5 | `t.$rowtime` (qualified system column) | CC exposes `$rowtime` as a built-in event-time system column on auto-mapped tables. It may be referenced unqualified or qualified with a table alias (e.g. `o.$rowtime`) — both work (verified live) | LOW | CC Flink project |
+| 6 | `ALTER TABLE t ADD WATERMARK ...` | Uses `MODIFY`, not `ADD`, for watermarks — and also for other one-time schema recipes: overriding key/value types on schemaless topics (`` ALTER TABLE t MODIFY (`key` STRING, `val` STRING) ``), setting schema context (`ALTER TABLE t SET ('value.format.schema-context' = ...)`), and adding metadata columns | MEDIUM | CC Flink project |
```
MartijnVisser, line 18: **Row 6 generalizes `MODIFY` over a recipe that actually needs `SET`.** The headline reads "Uses `MODIFY`, not `ADD`, for watermarks, and also for other one-time schema recipes", and then lists among those recipes: setting schema context via `ALTER TABLE t SET ('value.format.schema-context' = ...)`.

The inline example is right, but it uses `SET`, not `MODIFY`, because a table-option change is not a schema/column change. A reader who takes the row headline at face value could emit `ALTER TABLE t MODIFY ('value.format.schema-context' = ...)`, which is not valid syntax for a table option.

Suggest splitting the row, or rewording so the `MODIFY` claim covers only the schema/column recipes (watermarks, key/value type overrides, metadata columns) and the schema-context case is called out separately as `SET`.

skills/confluent-cloud-flink-sql/references/dialect-traps.md
```diff
@@ -0,0 +1,63 @@
+# Confluent Cloud vs Apache Flink SQL — Dialect Traps
+
+Single source of truth. Update here; per-project CLAUDE.md references this file.
+
+Reviewed by a Flink PMC / CC-for-Apache-Flink SME via live reproducers on a real CC compute pool. Two originally-listed traps were refuted by that testing and have been removed (see history below); do not re-add them without a live repro.
```
MartijnVisser, line 5: **Nit (bookkeeping):** this says "Two originally-listed traps were refuted by that testing and have been removed", which matches the two entries actually marked *refuted*. But the section it points to is headed "Removed traps (refuted by live testing, do not re-add)" and contains three bullets, the third being the `SESSION_START` entry marked *obsolete* rather than refuted. So the heading labels all three as refuted-by-live-testing while the intro counts two.

Either adjust the heading (for example "Removed traps (refuted or obsolete, do not re-add)") or move the obsolete entry out, so the count and the label line up. Cosmetic only.

MartijnVisser (commented):
Third pass. All four findings from my last review are confirmed fixed at this head: the `$rowtime AS alias` red flag is gone from SKILL.md, the CLI required-flag text now correctly lists `--sql`/`--compute-pool`/`--database`/`--environment` (and the sentence, table, and key-rules bullet all agree, which is the part that usually drifts), the LAG window is computed once instead of three times, and the redundant outer CAST is dropped. The trap-table count (22) also matches the actual 22 numbered rows. Thanks for the quick turnaround.

Three smaller items left, none of them blocking. Details inline.

One item I want to keep closed: an automated pass flagged the single-statement `KEY_SEARCH_AGG` + `CROSS JOIN UNNEST(search_results)` example again as possibly unbindable. I verified that empirically last round on a live CC pool (runtime flink-2.0.2553.0-v1): `EXPLAIN` on that exact combined statement COMPLETED, and a negative control substituting a bogus column name FAILED with "Column 'bogus_results' not found in any table". So `search_results` genuinely binds in the combined form. That example is correct as written and needs no change.

MartijnVisser (approved):
Approving. Every blocking item from the earlier rounds is fixed and verified at this head: the `$rowtime AS alias` red flag is gone from SKILL.md, the CLI required-flag guidance now correctly lists `--sql`/`--compute-pool`/`--database`/`--environment` with the sentence, table, and key-rules bullet all in agreement, the LAG window is computed once rather than three times, the redundant outer CAST is dropped, and the trap-table count matches its 22 rows.

The three comments I left on the previous pass are **non-blocking** and fine to pick up in a follow-up:

1. `cli-reference.md:107`, the eventual-consistency example gates on topic existence directly under a rule that says to gate on side-effects instead. This is the one I would most like to see land eventually, since it is the example an agent is most likely to copy, but it should not hold up the merge.
2. `dialect-traps.md:18`, row 6's `MODIFY` headline covers a recipe that actually needs `SET`.
3. `dialect-traps.md:5`, cosmetic count/label mismatch in the removed-traps section.

To restate one thing so it does not get re-opened by a future automated pass: the single-statement `KEY_SEARCH_AGG` + `CROSS JOIN UNNEST(search_results)` example is correct as written. I verified it on a live CC pool (runtime flink-2.0.2553.0-v1): `EXPLAIN` COMPLETED on that exact statement, and a negative control with a bogus column name failed with "Column 'bogus_results' not found in any table", so `search_results` genuinely binds in the combined form. No change needed there.

Nice work turning these rounds around quickly.

PR #55, opened 2026-08-06
confluent-skill-creator: fix compatibility visibility, script fallback, description quoting

leeec1993:
## Summary

Three fixes to `skills/confluent-skill-creator/SKILL.md`, sourced from the [Agent Skills Portability Review](https://confluentinc.atlassian.net/wiki/spaces/DTX/pages/5991203400/Agent+Skills+Portability+Review+agentskillimplementation.com+Findings) (agentskillimplementation.com benchmark findings, 2026-08-04). Both "Real gaps" from that review name `confluent-skill-creator` directly.

1. **`compatibility:` field is invisible to the model on Claude Code.** Per the benchmark's Claude Code report, this platform strips YAML frontmatter beyond `name`/`description` before injecting the file into the model — so the skill's own runtime dependencies were never actually surfaced. Added a `## Prerequisites` section to the body.

   While fixing this, found the dependency list itself was stale: the frontmatter said "confluent-kafka, fastavro, requests", but `scripts/requirements.txt` actually lists `confluent-kafka[avro,json,protobuf]`, `requests`, `jsonschema`, `python-dotenv` — no `fastavro` at all. Corrected in both the frontmatter and the new Prerequisites section, and pointed the latter at `requirements.txt` as source of truth so it can't drift silently again.

2. **Bundled scripts have no documented fallback if execution is blocked.** The benchmark's Claude Code test found script execution gets denied by the permission system, with no approval mechanism in headless/non-interactive sessions. `confluent-skill-creator` ships 5 scripts (`check_compute_pool.py`, `produce_data.py`, `consume_and_verify.py`, `register_schema.py`, `cleanup_resources.py`) with no recovery path documented. Added a manual-steps fallback for each, plus a requirement that skills *created by* this skill document the same fallback if they ship their own scripts.

3. **Inconsistent description quoting.** `description:` was unquoted, unlike 3 of the other 8 skills in the repo. Quoted for consistency and to avoid breaking on platforms with stricter YAML parsing (e.g. Antigravity) if a raw colon is ever introduced.

## Out of scope (follow-ups, not fixed here)

- **`CONTRIBUTING.md` scope-precedence documentation** and the **2-platform-vs-broader-portability roadmap decision** — both raised by the same review, but repo-level, not file-level fixes to this skill.
- **`scripts/requirements.txt` lists `python-dotenv`, but no script actually calls `load_dotenv()`** — all 5 scripts read credentials via plain `os.getenv()`, expecting the `.env` file already sourced into the shell. Checked PyPI metadata too: `python-dotenv` isn't pulled in transitively by anything else either. Looks like dead weight, but leaving `requirements.txt` untouched to keep this PR scoped to the portability-review fixes.

## Testing

Docs-only change (SKILL.md body + frontmatter). No script behavior changed.

### DIFF
--- a/skills/confluent-skill-creator/SKILL.md
+++ b/skills/confluent-skill-creator/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: confluent-skill-creator
-description: Create Confluent-specific skills for external users. Use this skill when users want to create, build, or author a new skill related to Confluent Cloud, Confluent Platform, Apache Kafka, WarpStream, Flink, Connectors, Schema Registry, Tableflow, CDC pipelines, or any Confluent product. Skills can be use-case focused (like data enrichment, CDC to Tableflow, stream processing workflows) or component-specific (like a Flink skill, Schema Registry skill, or Connector skill). Do NOT use this skill when users want to directly use Confluent products (e.g., build a pipeline, write a producer, deploy Flink SQL) — use the appropriate product-specific skill instead. This skill is specifically for creating new skills, not for using existing ones.
-compatibility: Requires Python 3.9+, confluent-kafka, fastavro, requests. Needs access to a Confluent environment (Cloud, Platform, local Docker, or WarpStream) for E2E testing.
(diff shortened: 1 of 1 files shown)


Cerchie (approved):
LGTM from a quick look, let's iterate as we use it

